PR #214, opened 2014-04-01
Cast always quantity to integer in calculatePrice

jerome-fix:
When we hide the quantity input in method BaseProductProvider::defineAddBasketForm($product, $formBuilder, true, $option), the quantity is send to string format (="1") and not as an integer (=1).

In this case an exception is thrown systematically.


OskarStark:
Closed by #243

PR #229, opened 2014-04-07
Update BaseProduct.php

loicolas:
Avoid Php Warning when product has no associated category.


OskarStark:
@loicolas could you add the tests to the PR?

Issue #258, opened 2014-05-26
Product variations list error


wbloszyk:
Method "enabledVariations" for object "Application\Sonata\ProductBundle\Entity\Travel" does not exist in SonataProductBundle:Travel:view_variations.html.twig at line 18

PR #260, opened 2014-05-27
fix product variations list bug

wbloszyk:
Fix for /shop/product/{slug}/{productId} router - sonata_product_view_variations
#258


OskarStark:
can you please execute `make cs` ? thank you

PR #273, opened 2014-08-16
Update view.html.twig

piron4o:
```php
DateTimeHelper::formatDateTime($datetime, $locale = null, $timezone = null, $dateType = null, $timeType = null);
```
third parameter must be `(string) $timezone`


src/InvoiceBundle/Resources/views/Invoice/view.html.twig
```diff
@@ -54,7 +54,7 @@ file that was distributed with this source code.
                 <table class="table table-bordered">
                     <tr>
                         <th>{% trans from 'SonataInvoiceBundle' %}sonata.invoice.references.date{% endtrans %}</th>
-                        <td>{{ invoice.createdAt|format_datetime(null, invoice.locale, date_time_size, date_time_size) }}</td>
+                        <td>{{ invoice.createdAt|format_datetime(null, invoice.locale, null,date_time_size, date_time_size) }}</td>
```
OskarStark: is this correct, that we use `date_time_size` twice?

PR #290, opened 2014-12-26
Add support for Cocur Slugify library

meyerbaptiste:
Like SonataPageBundle, we need to add support for Cocur Slugify library.


meyerbaptiste:
ping @rande

PR #297, opened 2015-01-15
Implement pagination to API Order

bdejacobet:
Implement API pagination through PageableManagerInterface with unit testing


src/OrderBundle/Entity/OrderManager.php
```diff
@@ -64,4 +66,44 @@ public function getOrder($orderId)
...
+    public function getPager(array $criteria, $page, $limit = 10, array $sort = array())
+    {
+        $query = $this->getRepository()
+            ->createQueryBuilder('o')
+            ->select('o');
+
+        $fields = $this->getEntityManager()->getClassMetadata($this->class)->getFieldNames();
+        foreach ($sort as $field => $direction) {
+            if (!in_array($field, $fields)) {
+                unset($sort[$field]);
```
amine26tz: throw exception there ;)

PR #302, opened 2015-01-26
Update OgonePayment.php

stphnpt:
In compareOrderToParams, comparison shall be made on value (and not on type) as $order->getCurrency() is now Object whereas $params['currency'] is String (currently throws a UnauthorizedHttpException)


rande:
Can you update unit test ?

stphnpt:
Sorry, I'm still not deep enough in this area to update payment component unit test ... I would certainly make mistakes !

PR #309, opened 2015-01-27
Fix interface compatibility

MrGreenStuff:


rande:
@MrGreenStuff is this change compatible with symfony 2.6 ?

sergiostrampelli:
Using Symfony 2.6, I have patched with:

    public function validateOneMainCategory(ExecutionContextInterface $context)

MrGreenStuff:
Ok for 2.6

rande:
Does that work for symfony 2.3 too ? (this is the minimun requirements in the composer.json)

Issue #293, opened 2015-01-12
Creation of product  "validateOneMainCategory()"


Mays:
Hello,

When i   add a  product in sonata admin i get this

Catchable Fatal Error: Argument 1 passed to Sonata\ProductBundle\Entity\BaseProduct::validateOneMainCategory() must be an instance of Symfony\Component\Validator\ExecutionContext, instance of Symfony\Component\Validator\Context\LegacyExecutionContext given in /media/mehdi/Data1/work/phpwork/cookbook/vendor/sonata-project/ecommerce/src/ProductBundle/Entity/BaseProduct.php line 949

grosenbaum:
ping @rande

grosenbaum:
It's corrected now. Could you check and close this issue ?

cbastienbaron:
indeed :+1:

cbastienbaron:
already merge : https://github.com/sonata-project/ecommerce/pull/314

branch dev-develop

jordisala1991:
Closing as it is already fixed

PR #314, opened 2015-01-29
change validator context of product entity

bdejacobet:
Fixing issue #293

(fix done by @utilisezlinux but there was a problem with his PR )

### DIFF
--- a/src/Component/Product/ProductInterface.php
+++ b/src/Component/Product/ProductInterface.php
@@ -14,7 +14,7 @@ namespace Sonata\Component\Product;
 use Sonata\MediaBundle\Model\GalleryInterface;
 use Sonata\MediaBundle\Model\MediaInterface;
 use Doctrine\Common\Collections\ArrayCollection;
-use Symfony\Component\Validator\Context\LegacyExecutionContext;
+use Symfony\Component\Validator\ExecutionContextInterface;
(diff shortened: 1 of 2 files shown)


rande:
@bdejacobet not sure it is the correct class: ``https://github.com/symfony/Validator/blob/master/Context/ExecutionContextInterface.php``

utilisezlinux:
According to https://github.com/symfony/Validator/tree/master/Context, LegacyExecutionContext.php and ExecutionContext.php are deprecated and to be suppressed in symfony 3.0.
The advice given by Bernhard Schussek <bschussek@gmail.com> is `not to instantiate or use this class. Code against {@link ExecutionContextInterface} instead.`
The fixing works for me with changes in`https://github.com/sonata-project/ecommerce/blob/master/src/ProductBundle/Entity/BaseProduct.php` and in `https://github.com/sonata-project/ecommerce/blob/master/src/Component/Product/ProductInterface.php`.

Should we use this fix temporarily or modify the call of validateOneMainCategory() to pass ExecutionContext instead of LegacyExecutionContext to it?

bdejacobet:
@rande change to Symfony\Component\Validator\ExecutionContextInterface

rande:
Thanks

PR #316, opened 2015-01-30
transform descriptions with formatter in API method postProductAction

bdejacobet:


rande:
You need to add composer self-update inside travis file

bdejacobet:
@rande travis file updated

rande:
Thanks

PR #324, opened 2015-02-16
Update validation of address form

bdejacobet:
Update validation of address form with to groups
* 'Default' : group used by the default form (used by API) where 'type' can not be null
* 'front' : group used by the basketBundle where type is not mandatory (it is defined before the save of the address with the appropriate value)


bdejacobet:
@rande PR updated

PR #340, opened 2015-06-07
Update OgonePayment

stphnpt:


stphnpt:
Type comparison (string / string) should now go through

PR #351, opened 2015-10-07
Remove "number" field from configureShowFields in BaseProductProvider

stphnpt:
there is no "number" property in BaseProduct whereas "priceIncludingVat" exists


Soullivaneuh:
* Tests are failing
* Please provide a better commit comment and PR title.

stphnpt:
When reading the Travis log (build could not complete Time-out), I am unable to understand what I can do to make it pass through. Would you be so kind as to explain it to me ?

PR #369, opened 2016-06-13
Adding supports for disabled products (SEO friendliness)

AymericWilke:
Hi,
When we disable a product and then go on the page of this product, it results in 404, which is pretty bad for indexation bots.
So I added the ability to show the page of disabled products, with a notice. The products still don't appear in the catalog, similar, etc pages, and you can't add them in your cart.

### DIFF
--- a/src/ProductBundle/Resources/translations/SonataProductBundle.en.xliff
+++ b/src/ProductBundle/Resources/translations/SonataProductBundle.en.xliff
@@ -306,6 +306,10 @@
                 <source>product_price</source>
                 <target>Price</target>
             </trans-unit>
+            <trans-unit id="product_not_available">
+                <source>product_not_available</source>
+                <target>This product is not available anymore.</target>
+            </trans-unit>
             <trans-unit id="product_variations">
                 <source>product_variations</source>
                 <target>Variations</target>
--- a/src/ProductBundle/Resources/views/Product/view.html.twig
+++ b/src/ProductBundle/Resources/views/Product/view.html.twig
@@ -69,10 +69,16 @@ file that was distributed with this source code.
                             {% block product_generic_properties %}
                                 <li class="list-group-item">
                                     {% block product_unit_price %}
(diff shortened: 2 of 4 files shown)


OskarStark:
Can you please add some screenshots?

src/ProductBundle/Resources/translations/SonataProductBundle.en.xliff
```diff
@@ -306,6 +306,10 @@
                 <source>product_price</source>
                 <target>Price</target>
             </trans-unit>
+            <trans-unit id="product_not_available">
```
OskarStark: please update the other translation files, too

src/ProductBundle/Resources/views/Product/view.html.twig
```diff
@@ -69,10 +69,16 @@ file that was distributed with this source code.
...
+                                        {% if product.enabled %}
+                                            <small class="list-group-item-text">
+                                                {{ 'product_price'|trans({}, "SonataProductBundle") }}:&nbsp;
+                                                <span itemprop="price">{{ sonata_product_price(product, currency, true)|number_format_currency(currency) }}</span>
+                                            </small><br/>
+                                        {% else %}
+                                            <small class="list-group-item-text">
+                                                <span>{{ 'product_not_available'|trans({}, "SonataProductBundle") }}</span>
+                                            </small><br/>
+                                        {% endif %}
```
greg0ire: The outer `small` tag does not change, you should put it outside of the condition.

src/ProductBundle/Resources/translations/SonataProductBundle.en.xliff
AymericWilke: Sorry… what other translation files ?

OskarStark: You are right, sorry i thought we have some other languages here 😄

PR #383, opened 2016-06-25
Remove internal test classes from autoloader

core23:
I am targetting this branch, because this is a major BC break for all tests.

Related to https://github.com/sonata-project/dev-kit/issues/179

## Changelog

```markdown
### Removed
 - internal test classes are now excluded from the autoloader
```

## Subject

Tests for internal components shouldn't be loaded in production or extended by other bundles.

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -4,3 +4,7 @@ UPGRADE FROM 1.x to 2.0
 ### Sonata dependencies
 
 Bump all Sonata dependencies to versions managed by the new management system.
+
+### Tests
+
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
I don't think we should make it on master because no one is supposed to use our test classes.

I just said to test internal relation before doing that: https://github.com/sonata-project/dev-kit/issues/179#issuecomment-227391842

core23:
> I don't think we should make it on master because no one is supposed to use our test classes.

Should I open this against the stable 2.x branch?

> I just said to test internal relation before doing that: sonata-project/dev-kit#179 (comment)

No other sonata bundle has a dependency to this, so we could merge this without any risk.

Soullivaneuh:
> No other sonata bundle has a dependency to this, so we could merge this without any risk.

I trust you. Ok for 2.x then.

PR #417, opened 2017-03-16
Use cocur-slugify for BaseProduct slug

ZRazor:
I am targeting this branch, because its BC.

## Changelog

```markdown
### Changed
- Changed `BaseProduct::setSlug` to use `cocur-slugify`
- Deprecated `BaseProduct::slugify`
```

## Subject

Fixes problem with cyrillic

### DIFF
--- a/src/ProductBundle/Entity/BaseProduct.php
+++ b/src/ProductBundle/Entity/BaseProduct.php
@@ -11,6 +11,7 @@
 
 namespace Sonata\ProductBundle\Entity;
 
+use Cocur\Slugify\Slugify;
 use Doctrine\Common\Collections\ArrayCollection;
 use Sonata\Component\Product\DeliveryInterface;
 use Sonata\Component\Product\PackageInterface;
@@ -245,7 +246,9 @@ abstract class BaseProduct implements ProductInterface
      */
     public function setSlug($slug)
     {
-        $this->slug = self::slugify(trim($slug));
+        $slug = Slugify::create()->slugify(trim($slug));
+
+        $this->slug = !empty($slug) ? $slug : 'n-a';
     }
 
     /**
@@ -845,38 +848,21 @@ abstract class BaseProduct implements ProductInterface
     }
 
     /**
-     * source : http://snipplr.com/view/22741/slugify-a-string-in-php/.
+     * NEXT_MAJOR: remove this method.
      *
      * @static
      *
      * @param  $text
      *
      * @return mixed|string
+     *
+     * @deprecated since 2.1, to be removed with 3.0
      */
     public static function slugify($text)
     {
-        // replace non letter or digits by -
-        $text = preg_replace('~[^\\pL\d]+~u', '-', $text);
+        $text = Slugify::create()->slugify(trim($text));
 
-        // trim
-        $text = trim($text, '-');
-
-        // transliterate
-        if (function_exists('iconv')) {
-            $text = iconv('utf-8', 'us-ascii//TRANSLIT', $text);
-        }
-
-        // lowercase
-        $text = strtolower($text);
-
-        // remove unwanted characters
-        $text = preg_replace('~[^-\w]+~', '', $text);
-
-        if (empty($text)) {
-            return 'n-a';
-        }
-
-        return $text;
+        return !empty($text) ? $text : 'n-a';
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
     ],
     "require" : {
         "php": "^5.3 || ^7.0",
+        "cocur/slugify": "^1.4 || ^2.0",
         "friendsofsymfony/comment-bundle": "^2.0",
         "friendsofsymfony/rest-bundle": "^1.1 || ^2.0",
         "friendsofsymfony/user-bundle": "^1.3",


src/ProductBundle/Entity/BaseProduct.php
```diff
@@ -855,22 +854,7 @@ public function isPriceIncludingVat()
...
-        if (function_exists('iconv')) {
-            $text = iconv('utf-8', 'us-ascii//TRANSLIT', $text);
-        }
-
-        // lowercase
-        $text = strtolower($text);
-
-        // remove unwanted characters
-        $text = preg_replace('~[^-\w]+~', '', $text);
+        $text = Slugify::create()->slugify($text);
```
greg0ire: DIP violation, please inject this.

greg0ire: Plus it's an entity, the entity shouldn't have a `slugify` method…

ZRazor:
Is it better now?

src/ProductBundle/Entity/BaseProduct.php
```diff
@@ -845,41 +848,6 @@ public function isPriceIncludingVat()
...
     /**
-     * source : http://snipplr.com/view/22741/slugify-a-string-in-php/.
-     *
-     * @static
-     *
-     * @param  $text
-     *
-     * @return mixed|string
-     */
-    public static function slugify($text)
```
OskarStark: removing this is a bc break

ZRazor: Then, I suppose it's better to keep `slugify`, even if its a bit ugly. Or I can target master branch.

Soullivaneuh: @ZRazor All removal must be deprecated on current stable before removing on master.

ZRazor: I can keep `setSlug` as it is now:
```php
$slug = Slugify::create()->slugify(trim($slug));

$this->slug = !empty($slug) ? $slug : 'n-a';
```
And keep changed `BaseProduct::slugify` with adding `@deprecated` to it

src/ProductBundle/Entity/BaseProduct.php
```diff
@@ -245,7 +246,9 @@ public function getSku()
      */
     public function setSlug($slug)
     {
-        $this->slug = self::slugify(trim($slug));
+        $slug = Slugify::create()->slugify(trim($slug));
```
greg0ire: The dependency is still not injected. It should probably be `setSlug($slug, Slugify slugifier)` (but simulated with `func_get_args()`. I'm not sure of the implications.

greg0ire: Anyway, maybe it's good enough like this… not that serious.

ZRazor:
Did some changes. Should be BC now, and dependency is injected.

src/ProductBundle/Entity/BaseProduct.php
```diff
@@ -243,9 +244,14 @@ public function getSku()
     /**
      * {@inheritdoc}
      */
-    public function setSlug($slug)
+    public function setSlug($slug, Slugify $slugifier = null)
```
greg0ire: When is `setSlug` called?

greg0ire: Better solutions imply events, like doctrine prePersist or preUpdate events, but not sure it will fit our use-case…

ZRazor: In `setName`

src/Component/Product/ProductInterface.php
```diff
@@ -49,9 +50,10 @@ public function getSku();
     /**
      * Set slug.
      *
-     * @param string $slug
+     * @param string  $slug
+     * @param Slugify $slugifier
      */
-    public function setSlug($slug);
+    public function setSlug($slug, Slugify $slugifier = null);
```
greg0ire: This is a BC-break…

greg0ire:
@ZRazor this one is hard… maybe have a look at how this is done usually? I think it's done from the outside most of the time (the slug is generated and then injected). Plus, it's not really part of the domain, it's more something convenient for getting a product in an HTTP context. In some projects, for instance, slugs are stored in a separate table, that has an `entity` column.

ZRazor:
@greg0ire For example: https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/Model/Tag.php#L152
I'm using this bundles in my project, and they have tonns of bugs. I have already overrided half of the package to get a working project. So I think we should just ignore DIP violation for now and at least deprecate `BaseProduct::slugify`
What do you think?

greg0ire:
@ZRazor , I agree! It will still be a big improvement.

OskarStark:
What about checking for the already native sluger from admin bundle (or is it in the core bundle ❓ ) if it is defined as a service and the auto inject it via a setter.

This could then be used if is set, if not, we keep the old logic in the method and throw a deprecation warning

greg0ire:
The problem is, how and when to inject it…

OskarStark:
> The problem is, how and when to inject it…

would do this via a compiler pass

ZRazor:
@OskarStark, do you mean `sonata.core.slugify.native`? It's deprecated, or i am wrong?

OskarStark:
> @OskarStark, do you mean sonata.core.slugify.native? It's deprecated, or i am wrong?

ZRazor:
Okay, I can inject it via a compiler pass and add a listener on `prePersist`, but at the current moment slug is updated in `setName` and maybe outside the bundle manually by user. So if we update slug in `prePersist` it can be a BC break, because someone can update it mannualy outside the bundle and doesnot change the name.

greg0ire:
> would do this via a compiler pass

That works with services, but this is an entity

ZRazor:
Does anyone have ideas what to do with this?

OskarStark:
> That works with services, but this is an entity

you are right, my fault

OskarStark:
@greg0ire so merge as it is?

greg0ire:
Yes please

PR #422, opened 2017-03-30
Remove deprecated Admin class, use AbstractAdmin

ZRazor:
## Changelog

```markdown
### Changed
- use `AbstractAdmin` instead of deprecated `Admin` class
```

## Note
In `ProductVariationAdmin` `Admin` is not needed, because it extends `ProductAdmin`


OskarStark:
I updated your changelog, will merge if green

PR #426, opened 2017-04-02
Increase core-bundle dependency to ^3.0.2

ZRazor:
I am targeting this branch, because this is pedantic.

## Subject

Needed for tests passing


jordisala1991:
Thanks @ZRazor

PR #429, opened 2017-04-03
Use parameters for forms classes in BasketBundle and CustomerBundle

ZRazor:
I am targeting this branch, because it's BC.

## Changelog

```markdown
### Added
- Added parameters for forms classes in `BasketBundle` and `CustomerBundle`
```

## Subject

These forms often need to be overrided to customize them.


jordisala1991:
Not sure if we allow those kind of changes. @greg0ire has a nice article that talks about it, I couldn't find it.

Not sure if this can be done in another way like defining those parameters on the config.yml using Configuration and Extension classes?

greg0ire:
We indeed don't. We keep the API closed until an extension point is proven to be needed

greg0ire:
For the article, I'm on my phone, sorry

greg0ire:
Here are the articles

- http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html
- https://ocramius.github.io/blog/when-to-declare-classes-final/

ZRazor:
I'm closing this.
We should do it (configuring address fields) another way

PR #432, opened 2017-04-10
Add missing basket property in AddressType

ZRazor:
I am targeting this branch, because this is a bug fix.

## Changelog

```markdown
### Added
- Added missing `basket` property in `AddressType`
```

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -1,6 +1,13 @@
 UPGRADE 2.x
 ===========
 
+UPGRADE FROM 2.1.1 to 2.x
+=========================
+
+### Sonata\CustomerBundle\Form\Type\AddressType
+
+If you redefined this class, note that missing `basket` property was added, which means it is no longer public.
+
 UPGRADE FROM 2.0 to 2.1
 =======================
 
--- a/src/CustomerBundle/Form/Type/AddressType.php
+++ b/src/CustomerBundle/Form/Type/AddressType.php
@@ -40,6 +40,11 @@ class AddressType extends AbstractType
     protected $name;
 
     /**
+     * @var BasketInterface
+     */
+    protected $basket;
+
+    /**
      * @param string          $class  A class to apply getter
      * @param string          $getter A getter method name
      * @param string          $name   A form type name


src/CustomerBundle/Form/Type/AddressType.php
```diff
@@ -40,6 +40,11 @@ class AddressType extends AbstractType
     protected $name;
 
     /**
+     * @var BasketInterface
+     */
+    protected $basket;
+
+    /**
```
greg0ire: @sonata-project/contributors technically, this is a BC-break, but I feel we should consider it patch nonetheless.

jordisala1991: Maybe add an upgrade note just for anyone using directly AddressType->basket, or redefining this class?

IMO it should be a patch

ZRazor: I added a note, is it what you meant?

greg0ire: I think he means editing https://github.com/sonata-project/ecommerce/blob/2.x/UPGRADE-2.x.md

greg0ire: Which you did :P perfect!

UPGRADE-2.x.md
```diff
@@ -9,3 +9,10 @@ UPGRADE FROM 2.0 to 2.1
 All files under the ``Tests`` directory are now correctly handled as internal test classes. 
 You can't extend them anymore, because they are only loaded when running internal tests. 
 More information can be found in the [composer docs](https://getcomposer.org/doc/04-schema.md#autoload-dev).
+
+UPGRADE FROM 2.1.1 to 2.1.2
+===========================
+
+### Sonata\CustomerBundle\Form\Type\AddressType
+
+If you redefined this class, note that missing `basket` property was added.
```
greg0ire: "… which means it is no longer public"

UPGRADE-2.x.md
```diff
@@ -9,3 +9,10 @@ UPGRADE FROM 2.0 to 2.1
 All files under the ``Tests`` directory are now correctly handled as internal test classes. 
 You can't extend them anymore, because they are only loaded when running internal tests. 
 More information can be found in the [composer docs](https://getcomposer.org/doc/04-schema.md#autoload-dev).
+
+UPGRADE FROM 2.1.1 to 2.1.2
```
jordisala1991: I think this has to be placed on top with a message like:

UPGRADE FROM 2.1.1 to 2.x

And when we do a release, we change it to the new version (it could be 2.1.2 or 2.2.0)

ZRazor: Done

jordisala1991:
Thanks @ZRazor :)

PR #441, opened 2017-05-02
Symfony 3 support

Filoz:
I am targeting this branch, because this will introduce BC Break.

## Changelog
```markdown
- use `TokenStorageInterface` instead of deprecated `SecurityContextInterface`
```

## To do
- [ ] Check if other part need to be updated
- [ ] Update the tests

## Subject

I started to update the Sonata Ecommerce to support Symfony >= 3


jordisala1991:
Hi,

First of all, thanks for contributing this :)

I think this can be done BC, and it should IMO (given your current changes).

Keep in mind:

- Need to support SF 2.3
- You can't add mandatory params on methods (even construct)
  - For your case you can remove type hinting and use something like this: 
      https://github.com/sonata-project/SonataUserBundle/blob/3.x/Block/AccountBlockService.php#L40
      https://github.com/sonata-project/SonataUserBundle/blob/3.x/DependencyInjection/SonataUserExtension.php#L79
- For the factory something like this: 
       https://github.com/sonata-project/SonataUserBundle/blob/3.x/DependencyInjection/SonataUserExtension.php#L90

We can help you changing all of this if you want, just ask. @greg0ire , @ZRazor or myself will answer for sure.

Next steps:

1. Rebase on 2.x
2. Try to reduce the number of commits (easier to review)

Filoz:
Hi, 
thank you!

so... I need to  make another pull request with the the edited code (according to the links you gave me)?
Or I can continue with this pull request?

The target branch is 2 or master?

thank you

jordisala1991:
target branch is 2.x, continue on this PR please 👍

Filoz:
In symfony 3 is not possible to call $this->getRequest() in controller  (for example https://github.com/sonata-project/ecommerce/blob/2.x/src/ProductBundle/Controller/CatalogController.php#L34 )

insted is necessary to add Request $request as argument of the controller.
Can I add the argument or I need some different solution to not add BC break?

greg0ire:
@Filoz please have a look at the controllers of the SonataAdminBundle, we solved the issue by writing our own `getRequest` method : https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Controller/CRUDController.php#L927-L934

Filoz:
I think I had rebased on 2.x branch....can you check if it's ok?

if you say ok, I will continue with the edit... 
Otherwise I can delete this pull request and start from a new one ...

ZRazor:
There are a lot of things to do to make these bundles compatiable with SF 3.0
Take a look at:
`$form->isBound()` in controllers. This is deprecated in 2.7 and removed in 3.0.
`Symfony\Component\Validator\ExecutionContextInterface` - should be replaced with `Symfony\Component\Validator\Context\ExecutionContextInterface`
`$this->context->addViolationAt` - has relation to previous one.
`$this->getRequest()`
`use Symfony\Component\Form\Extension\Core\ChoiceList\SimpleChoiceList` and `choice_list` option - need to be replaced with `choice_loader`

Filoz:
Yes,  you are absolute right...
this pull request is just started, there are many things to check and update.

But I started with the wrong branch and if you agree, I would like to delete this pull request and start a new one

jordisala1991:
you need to rebase on 2.x only, I already changed this PR base

ZRazor:
@Filoz, I suggest you to make these bundles fully compatable with SF 2.8 and keep BC with 2.3 by small PR's in 2.x branch. Than you can drop BC code and add some stuff for 3.0 in master branch.

Filoz:
I'm sorry, but I made a really big mistake, I'm not so used to git...
I really prefer to delete this pull request and start from scratch with branch 2.x ...

please, can I close this pull request and delete patch-1 branch?

jordisala1991:
Yes you can do it

greg0ire:
If you fix your local branch and use `git push --force`, it will be equivalent to starting from scratch, and we will not loose the discussion.

greg0ire:
The last commit, which is a merge commit, indicate that you did not use the `--force` option. You can start over by resetting to the previous commit, and then force push :

- `git reset --hard f176f6a`
- `git push --force`

src/Component/Customer/CustomerSelector.php
```diff
@@ -29,26 +30,33 @@ class CustomerSelector implements CustomerSelectorInterface
...
 
     /**
-     * @var \Symfony\Component\Security\Core\SecurityContextInterface
+     * @var \Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface
      */
-    protected $securityContext;
+    protected $tokenStorage;
+    
+    /**
+     * @var \Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface
```
OskarStark: please do not use the FQCN here

src/Component/Customer/CustomerSelector.php
```diff
@@ -29,26 +30,33 @@ class CustomerSelector implements CustomerSelectorInterface
...
-     * @param SecurityContextInterface $securityContext
-     * @param LocaleDetectorInterface  $localeDetector
+     * @param CustomerManagerInterface      $customerManager
+     * @param SessionInterface              $session
+     * @param TokenStorageInterface         $tokenStorage
+     * @param AuthorizationCheckerInterface $authorizationChecker
+     * @param LocaleDetectorInterface       $localeDetector
      */
-    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, SecurityContextInterface $securityContext, LocaleDetectorInterface $localeDetector)
+    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, TokenStorageInterface $tokenStorage, AuthorizationCheckerInterface $authorizationChecker, LocaleDetectorInterface $localeDetector)
```
OskarStark: this is a BC break, please keep BC

src/OrderBundle/Block/RecentOrdersBlockService.php
```diff
@@ -13,22 +13,22 @@
...
 use Symfony\Component\OptionsResolver\OptionsResolver;
-use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  * @author Hugo Briand <briand@ekino.com>
  */
-class RecentOrdersBlockService extends BaseBlockService
+class RecentOrdersBlockService extends AbstractBlockService
```
OskarStark: @greg0ire should this be done in a separate Commit/PR?

greg0ire: Yes, it does not have to do with sf 3

greg0ire:
@Filoz I made a similar PR for the SonataPageBundle, with maybe some fixes you could copy : https://github.com/sonata-project/SonataPageBundle/pull/810

greg0ire:
Seems like you got git back on track right? Don't hesitate to ask for help if you think you do wrong things again, it's almost always possible to repair that.

src/Component/Customer/CustomerSelector.php
Filoz: yes, I'm editing it according to what jordisala1991 said in https://github.com/sonata-project/ecommerce/pull/441#issuecomment-299504486 (I will push after some hours)

I have to add a parameter $authorizationChecker (without type hinting) as was done in https://github.com/sonata-project/SonataUserBundle/blob/6f37cdf9dc57c0c71ba3b01a1856dfd6f9c686fc/Security/EditableRolesBuilder.php  or you have another solution?

Filoz:
Thankyou @greg0ire !

src/Component/Customer/CustomerSelector.php
```diff
@@ -29,26 +31,48 @@ class CustomerSelector implements CustomerSelectorInterface
...
-     * @param LocaleDetectorInterface  $localeDetector
+     * NEXT_MAJOR: Go back to type hinting check when bumping requirements to SF 2.6+.
+     *
+     * @param CustomerManagerInterface                          $customerManager
+     * @param SessionInterface                                  $session
+     * @param TokenStorageInterface|SecurityContextInterface    $tokenStorage
+     * @param LocaleDetectorInterface                           $localeDetector
      */
-    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, SecurityContextInterface $securityContext, LocaleDetectorInterface $localeDetector)
+    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, $tokenStorage, $authorizationChecker, LocaleDetectorInterface $localeDetector)
```
Filoz: I added a BC break here the same as was done in https://github.com/sonata-project/SonataUserBundle/blob/6f37cdf9dc57c0c71ba3b01a1856dfd6f9c686fc/Security/EditableRolesBuilder.php
If you have another solution to not add a BC break I will fix

Filoz:
I added a BC break here (https://github.com/sonata-project/ecommerce/pull/441#discussion-diff-116762866L47)  the same as was done in https://github.com/sonata-project/SonataUserBundle/blob/6f37cdf9dc57c0c71ba3b01a1856dfd6f9c686fc/Security/EditableRolesBuilder.php
If you have another solution to not add a BC break I will fix

src/BasketBundle/Form/PaymentType.php
```diff
@@ -99,7 +100,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
         $sub = $builder->create('paymentMethod', $choiceType, array(
             'expanded' => true,
-            'choice_list' => new SimpleChoiceList($choices),
+            'choice_list' => function (Options $opts, $previousValue) use ($choices) {
```
ZRazor: This wont work, choice_list is removed in 3.0

Filoz: You are right... is it ok if I do something like:        
 if (interface_exists('Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface')) { 
//choice_loader
}else{
//choice_list
}

ZRazor: You can do it this way: https://github.com/sonata-project/SonataClassificationBundle/pull/280
Also I made a typo there ;) so check this too https://github.com/sonata-project/SonataClassificationBundle/pull/284

Filoz:
Hi, I need yours feedback:

I spent many time trying to understand why, every time I go to delivery step it can't find any delivery method for the location I set.

Now I understand why! in https://github.com/sonata-project/ecommerce/blob/2.x/src/Component/Form/Type/DeliveryChoiceType.php#L65 , choices array must be "flipped".
Otherwise it save in the db the 'name' instead the 'code' of the delivery method.

I checked how was did in [Sonata Admin](https://github.com/sonata-project/SonataAdminBundle/blob/025e8e1092ae80d305dcbeb817efb17e4ee7a4d2/Form/Type/Filter/DateTimeRangeType.php#L67) to fix that, and this is how i would like to change the linked code:

```
    public function configureOptions(OptionsResolver $resolver)
    {
        $choices = array();

        foreach ($this->pool->getMethods() as $code => $instance) {
            $choices[$instance->getName()] = $code ;
        }

        // NEXT_MAJOR: Remove (when requirement of Symfony is >= 2.7)
        if (!method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
            $choices = array_flip($choices);
        } else {
            // NEXT_MAJOR: Remove (when requirement of Symfony is >= 3.0)
            if (method_exists('Symfony\Component\Form\FormTypeInterface', 'setDefaultOptions')) {
                $choiceOptions['choices_as_values'] = true;
            }
        }

        $resolver->setDefaults(array(
            'choices' => $choices,
        ));
    }
```

is it ok for you?
**note:** I changes `$name` to `$code` because it was confusing; `$name` refers to the code related to the delivery method and not to its name

src/BasketBundle/Form/PaymentType.php
Filoz: Tank you, I will check!

Filoz: @ZRazor I checked you code...
do you think that I need to create a 'ChoiceLoader' for every Type (for example: DeliveryChoiceLoader', ...) or I can create one that works for all ? because the code will be always the same... 
or, another solution is to create an abstract class with the code and many class as many loader I need...

Filoz: There is something that I don't understand;
If I adapt you code to my situation like this:
```
if (!interface_exists('Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface')) {
            $sub = $builder->create('deliveryMethod', $choiceType, array(
                'expanded' => true,
                'choice_list' => new SimpleChoiceList($choices),
            ));
        }else {
            $sub = $builder->create('deliveryMethod', $choiceType, array(
                'expanded' => true,
                'choice_loader' => function (Options $opts, $previousValue) use ($choices) {
                    return new DeliveryChoiceLoader(array_flip($choices));
                },
            ));
        }
```
I get the following error:
`The option "choice_loader" with value Closure is expected to be of type "null" or "Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface", but is of type "Closure".`

instead, if I change it like this:
```
if (!interface_exists('Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface')) {
            $sub = $builder->create('deliveryMethod', $choiceType, array(
                'expanded' => true,
                'choice_list' => new SimpleChoiceList($choices),
            ));
        }else {
            $sub = $builder->create('deliveryMethod', $choiceType, array(
                'expanded' => true,
                'choice_loader' => new DeliveryChoiceLoader(array_flip($choices))
            ));
        }
```

it seems to be working

greg0ire:
> is it ok for you?

Why wouldn't it be ok? We accepted it in Sonata Admin ;)

Filoz:
Yes, but I preferred to ask your opinion before starting to change every occurrence of 'choices' with that code :)

greg0ire:
Oh right, I thought there was something special in your implementation, but then yeah, this is the way to solve it.

Filoz:
I can't understand a thing done in SonataCore: https://github.com/sonata-project/SonataCoreBundle/blob/d0519eb15529a1cbef326e8eea9982b59ad162d7/Form/Type/BaseStatusType.php#L46

This 4 argument of the constructor define if the list must be flipped or not but...
I can't find  any definition (xml or DI Extension) where the 4 argument is filled; this happen in Sonata Ecommerce and Sonata User too.

in particular if go to admin to edit an invoice, the status list is wrong (value and key should switch each other).
image

https://github.com/sonata-project/ecommerce/blob/2.x/src/InvoiceBundle/Admin/InvoiceAdmin.php#L73_

Filoz:
@greg0ire , I'm sorry for tagging you... but I'm a little confused, because I don't know weel every Sonata bundle.

Can you give a look to this: https://github.com/sonata-project/ecommerce/pull/441#issuecomment-302759503 ?

greg0ire:
You should throw an exception inside the constructor, that way you will see where the object is build, and where there should be a fourth argument.

Filoz:
@greg0ire  Yes, I did exactly what you described... but in Sonata User I can't find flip = true, I always find flip = false.
but... let's talk about Ecommerce:

if I add the 4 argument here https://github.com/sonata-project/ecommerce/blob/2.x/src/ProductBundle/Resources/config/form.xml#L4
like this <argument \> and I fill it in the DI ?
can it be an acceptable solution? it souldn't add BC Break...

greg0ire:
Yes, get the definition, and conditionally add (or replace) the fourth argument.

src/OrderBundle/Block/RecentOrdersBlockService.php
Filoz: Ok, I restored AbstractBlockService in latest commit

src/Component/Customer/CustomerSelector.php
Filoz: fixed

jordisala1991:
This can be closed, master branch is already compatible with SF3 and 4

PR #443, opened 2017-05-18
remove MopaBootrap dependency

### LINKED ISSUE #442: is MopaBootstrap Required?
In https://github.com/sonata-project/ecommerce/blob/7e778d57d1229df3f73b453404f9e7360b0ca923/src/BasketBundle/Resources/views/Form/label.html.twig#L1 and https://github.com/sonata-project/ecommerce/blob/7e778d57d1229df3f73b453404f9e7360b0ca923/src/CustomerBundle/Resources/views/Form/label.html.twig#L1

the twig file extends MopaBootstrapBundle that is not a required package.
and without MopaBoostrapBundle this is the result

`Unable to find template "MopaBootstrapBundle:Form:fields.html.twig" `

Filoz:
I am targeting this branch, because MopaBoostrap is not a dependency and should be removed.

Closes #442 

## Changelog

```markdown
### Fixed
- usage of `MopaBootstrapBundle`
```

## Subject

In different page there was a reference to MopaBootstrap bundle that is not a required library so it failed to load returning an exception (see #442 ).
Removing MopaBootstraBundle references make them work.

### DIFF
--- a/src/BasketBundle/Resources/views/Basket/delivery_address_step.html.twig
+++ b/src/BasketBundle/Resources/views/Basket/delivery_address_step.html.twig
@@ -17,8 +17,6 @@ file that was distributed with this source code.
 
 {% include 'SonataBasketBundle:Basket:stepper.html.twig' with {step: 'delivery'} %}
 
-{% form_theme form 'SonataBasketBundle:Form:label.html.twig' %}
-
 {{ form_start(form, {'attr': {'role': 'form', 'class': 'form-horizontal'}}) }}
     {{ form_errors(form) }}
 
--- a/src/BasketBundle/Resources/views/Basket/payment_address_step.html.twig
+++ b/src/BasketBundle/Resources/views/Basket/payment_address_step.html.twig
@@ -17,8 +17,6 @@ file that was distributed with this source code.
 
 {% include 'SonataBasketBundle:Basket:stepper.html.twig' with {step: 'billing'} %}
 
-{% form_theme form 'SonataBasketBundle:Form:label.html.twig' %}
-
 {{ form_start(form, {'attr': {'class': 'form-horizontal'}}) }}
 {{ form_errors(form) }}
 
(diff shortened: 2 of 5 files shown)


greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. More on http://keepachangelog.com

greg0ire:
Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

More details in CONTRIBUTING.md

Filoz:
Hi, thank you for you feedback...
So... I have to edit the commit subject with `git commit --amend` then make a `git push --force`? 

This commit shouldn't be a BC Break because the template file ad usually overwritten  because of the message inserted in every page
`{% sonata_template_box '....template. Feel free to override it.' %}`

greg0ire:
>  So... I have to edit the commit subject with git commit --amend then make a git push --force?

That will only amend the last commit, but we can still squash and use that new commit message afterwards.

Read ./CONTRIBUTING.md

OskarStark:
@greg0ire I marked this as a patch and adjusted the changelog, are you fine with it?

Otherwise, if it is no `patch`and *removed` instead it could not go to `2.x` IMO

greg0ire:
I'm fine with it.

Filoz:
Thank you @OskarStark !
@greg0ire ,  I changed the latest commit... I hope it's ok

greg0ire:
The *what* part is far better @Filoz, but the *why* is still missing.

Filoz:
I update the latest commit.

I'm really sorry, I don't have a good English and it is difficult for me to write... but I will try to improve my commit subject and description in the future :)
I really want to make this bundle works again in new version of Symfony...and maybe (in the future) adding some new feature...

greg0ire:
> I'm really sorry, I don't have a good English and it is difficult for me to write

Please write bad english over no english at all though :P We can fix the commit grammar mistakes if any just before merging.

>  I really want to make this bundle works again in new version of Symfony...and maybe (in the future) adding some new feature...

Thanks a lot, I'm glad to hear that!

Filoz:
Yes, the first thing I need to add when i will finish with Symfony 3 is the "digital product"/"downloadable product" support...

in particular, an easy way to remove delivery step if only digital products are selected

greg0ire:
Thanks @Filoz !

PR #484, opened 2017-12-03
Fix usage of deprecated methods in controllers

ZRazor:
I am targeting this branch, because it's BC.

## Changelog

```markdown
### Fixed
- Fixed usage of deprecated methods in controllers (Improve compatibility with SF 3)
```

### DIFF
--- a/src/BasketBundle/Controller/BasketController.php
+++ b/src/BasketBundle/Controller/BasketController.php
@@ -21,6 +21,8 @@ use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\DomCrawler\Form;
 use Symfony\Component\Form\Extension\Validator\ViolationMapper\ViolationMapper;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Intl\Intl;
@@ -36,7 +38,7 @@ class BasketController extends Controller
      *
      * @param Form $form
      *
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return Response
      */
     public function indexAction($form = null)
     {
@@ -45,7 +47,7 @@ class BasketController extends Controller
         ]);
 
         // always validate the basket
-        if (!$form->isBound()) {
+        if (!$form->isSubmitted()) {
             if ($violations = $this->get('validator')->validate($form)) {
                 $violationMapper = new ViolationMapper();
                 foreach ($violations as $violation) {
@@ -67,12 +69,12 @@ class BasketController extends Controller
     /**
      * Update basket form rendering & saving.
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return RedirectResponse|Response
      */
     public function updateAction()
     {
         $form = $this->createForm(BasketType::class, $this->get('sonata.basket'), ['validation_groups' => ['elements']]);
-        $form->handleRequest($this->get('request'));
+        $form->handleRequest($this->getCurrentRequest());
 
         if ($form->isValid()) {
             $basket = $form->getData();
@@ -96,11 +98,11 @@ class BasketController extends Controller
      * @throws MethodNotAllowedException
      * @throws NotFoundHttpException
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse|\Symfony\Component\HttpFoundation\Response
+     * @return RedirectResponse|Response
      */
     public function addProductAction()
     {
-        $request = $this->get('request');
+        $request = $this->getCurrentRequest();
         $params = $request->get('add_basket');
 
         if ('POST' != $request->getMethod()) {
@@ -170,7 +172,7 @@ class BasketController extends Controller
     /**
      * Resets (empties) the basket.
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return RedirectResponse
      */
     public function resetAction()
     {
@@ -182,7 +184,7 @@ class BasketController extends Controller
     /**
      * Displays a header preview of the basket.
      *
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return Response
      */
     public function headerPreviewAction()
     {
@@ -195,7 +197,7 @@ class BasketController extends Controller
      * Order process step 1: retrieve the customer associated with the logged in user if there's one
      * or create an empty customer and put it in the basket.
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return RedirectResponse
      */
     public function authenticationStepAction()
     {
@@ -214,10 +216,11 @@ class BasketController extends Controller
      *
      * @throws HttpException
      *
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse|\Symfony\Component\HttpFoundation\Response
+     * @return RedirectResponse|Response
      */
     public function paymentStepAction()
     {
+        $request = $this->getCurrentRequest();
         $basket = $this->get('sonata.basket');
 
         if (0 == $basket->countBasketElements()) {
@@ -241,8 +244,8 @@ class BasketController extends Controller
             'validation_groups' => ['delivery'],
         ]);
 
-        if ('POST' == $this->get('request')->getMethod()) {
-            $form->handleRequest($this->get('request'));
+        if ('POST' == $request->getMethod()) {
+            $form->handleRequest($request);
 
             if ($form->isValid()) {
                 // save the basket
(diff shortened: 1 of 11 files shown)


src/ProductBundle/Controller/ProductVariationAdminController.php
```diff
@@ -36,8 +41,15 @@ public function createAction(Request $request = null)
             throw new AccessDeniedException();
         }
 
+        // NEXT_MAJOR: Keep FQCN when bumping Symfony requirement to 2.8+.
+        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
```
greg0ire: Please use the `::class` notation as much as you can.

ZRazor: Is it better now?

src/ProductBundle/Controller/ProductVariationAdminController.php
```diff
@@ -36,8 +42,15 @@ public function createAction(Request $request = null)
             throw new AccessDeniedException();
         }
 
+        // NEXT_MAJOR: Keep FQCN when bumping Symfony requirement to 2.8+.
+        if (method_exists(\Symfony\Component\Form\AbstractType::class, 'getBlockPrefix')) {
```
greg0ire: It is better, but please use a use statement too

ZRazor: Done

greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

Also, in your case, end users do not care about phpdoc changes

src/BasketBundle/Controller/BasketController.php
```diff
@@ -522,7 +524,19 @@ public function finalReviewStepAction()
...
+            'tac_error' => 'POST' == $this->getRequest()->getMethod(),
         ]);
     }
+
+    /**
+     * @return Request
+     */
+    public function getRequest()
+    {
+        if ($this->container->has('request_stack')) {
```
jordisala1991: I think we can safely rewrite those to always return `return $this->container->get('request_stack')->getCurrentRequest();`

src/BasketBundle/Controller/BasketController.php
```diff
@@ -522,7 +524,19 @@ public function finalReviewStepAction()
...
             'basket' => $basket,
-            'tac_error' => 'POST' == $this->get('request')->getMethod(),
+            'tac_error' => 'POST' == $this->getRequest()->getMethod(),
         ]);
     }
+
+    /**
+     * @return Request
+     */
+    public function getRequest()
```
jordisala1991: Can't we make those private?

src/ProductBundle/Controller/ProductVariationAdminController.php
```diff
@@ -36,8 +43,15 @@ public function createAction(Request $request = null)
...
+        // NEXT_MAJOR: Keep FQCN when bumping Symfony requirement to 2.8+.
+        if (method_exists(AbstractType::class, 'getBlockPrefix')) {
+            $integerType = IntegerType::class;
+        } else {
+            $integerType = 'integer';
+        }
+
         $form = $this->createFormBuilder(null, [])
-            ->add('number', 'integer', [
+            ->add('number', $integerType, [
```
jordisala1991: Remove the `if` and make this: `IntegerType::class`

src/BasketBundle/Controller/BasketController.php
ZRazor: No, this method is public in parent class in SF 2.8. But we can rename it, make private and replace all usages of getRequest().

src/ProductBundle/Controller/ProductVariationAdminController.php
ZRazor: Okay, SF < 2.8 is dropped, so no more this stuff with `if` anywhere ;)

jordisala1991: I am preparing a PR to fully replace those, If you want, don't change this one

src/BasketBundle/Controller/BasketController.php
jordisala1991: Not sure if it is a good idea to name it like the parent, because on SF 3.4 I think it is removed or final?

Maybe give it another name like: `getCurrentRequest`?

jordisala1991:
Maybe the changelog could be something like, Improve compatibility with SF 3?

src/BasketBundle/Controller/BasketController.php
ZRazor: Done

src/BasketBundle/Controller/BasketController.php
ZRazor: Done

src/BasketBundle/Controller/BasketController.php
```diff
@@ -47,7 +49,7 @@ public function indexAction($form = null)
         ]);
 
         // always validate the basket
-        if (!$form->isBound()) {
+        if (!$form->isSubmitted()) {
```
jordisala1991: This change is not related to this PR, is this because of Sf3 compat?

ZRazor: I have already fixed these deprecations in previous PR, but missed this one. And yes - this is for SF 3.

jordisala1991:
Can you simplify some calls? There are cases where `getCurrentRequest` gets called like 5 or 6 times, would be better to store `$request` in a variable and use it everywhere.

ZRazor:
@jordisala1991 Done.

src/ProductBundle/Controller/ProductController.php
```diff
@@ -276,4 +279,12 @@ public function variationToProductAction($productId, $slug)
 
         return $response;
     }
+
+    /**
+     * @return Request
+     */
+    private function getCurrentRequest()
```
OskarStark: can we add this to the CRUDController? Or create an BaseEcommerceCRUDController to avoid this duplication?

ZRazor: mb use trait?

OskarStark: sounds good 👍 

WDYT @greg0ire ?

greg0ire: It's something that will be removed so a trait kinda makes sense here, because it's arguably easier to remove, and allows to keep that method private.

OskarStark: we should add a `NEXT_MAJOR` above the usage.

What about a deprecation in the trait when it will be removed?

ZRazor: I made a try.
We'll remove `CurrentRequestTrait` in 3.x and inject Request $request in actions parameters.

greg0ire:
Please rebase to get rid of the merge commit.

ZRazor:
@greg0ire Done

src/Component/Controller/CurrentRequestTrait.php
```diff
@@ -0,0 +1,29 @@
...
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace Sonata\Component\Controller;
+
+use Symfony\Component\HttpFoundation\Request;
+/**
+ * NEXT_MAJOR: Remove this class
```
greg0ire: It's not a class

src/Component/Controller/CurrentRequestTrait.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @deprecated Class to be removed in 3.x
```
greg0ire: Still not a class :P

ZRazor: My bad ;)

src/Component/Controller/CurrentRequestTrait.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Symfony\Component\HttpFoundation\Request;
+/**
+ * NEXT_MAJOR: Remove this trait
+ *
+ * @deprecated Trait to be removed in 3.x
+ *
+ * @author Anton Zlotnikov <exp.razor@gmail.com>
+ */
+trait CurrentRequestTrait
```
core23: 👎 for this trait, because it could be used by other people => we couldn't drop this on a stable branch

jordisala1991:
Can we get an agreement here? to not make @ZRazor implement something and then go back and so on?

Do we need a trait here? IMO it reduces some code needed, but I agree on the fact that I prefer to add code that we can easily remove if we need to, and not be forced to keep it because BC compat.

My opinion is to go back to private methods. WDYT?

ZRazor:
So, trait or methods in each controller?
Trait is deprecated, so we warn users from the beginning that it will be removed. Anyway we'll soon inject Request into actions params, and there is no big difference between removing a method at the end of 6 controllers or removing trait and it use statements.

ZRazor:
@core23 Done

jordisala1991:
Can you do a rebase? we merged a PR to improve our lint process

ZRazor:
Oh this rebasing without merge commit 🌚 
@jordisala1991, Done.

greg0ire:
Thanks @ZRazor !

Issue #404, opened 2017-02-03
Default precision in BaseOrder and BaseOrderElement


ZRazor:
## Subject

`BaseOrderElement` decimal fields are mapped without specific precision
```xml
<field name="unitPriceExcl" type="decimal" column="unit_price_excl" scale="4"/>
<field name="unitPriceInc" type="decimal" column="unit_price_inc" scale="4"/>
<field name="price" type="decimal" column="price" scale="4"/>
<field name="vatRate" type="decimal" column="vat_rate" scale="4"/>
<field name="designation" type="string" column="designation" length="255"/>
```
Same situation for `BaseOrder`
```xml
<field name="totalInc" nullable="true" type="decimal" column="total_inc" scale="4"/>
<field name="totalExcl" nullable="true" type="decimal" column="total_excl" scale="4"/>
<field name="deliveryCost" nullable="true" type="decimal" column="delivery_cost" scale="4"/>
<field name="deliveryVat" nullable="true" type="decimal" column="delivery_vat" scale="4"/>
```

By default in mysql and postgresql precision for decimal fields is equal to 10. So if we have an order or an order element with totalPrice over than 10^6 it will fail while inserting into the database.

## Steps to reproduce

Create a product with price 500 000
Add it to the basket
Set quantity to 20
Try to make an order

## Expected results
Created order

## Actual results
Database inserting fail

## Current fix
Override attributes in `Application\Sonata\OrderBundle\Entity\Order`
```php
/**
 * Class Order
 * @ORM\Entity()
 * @ORM\Table(name="order__order")
 * @ORM\AttributeOverrides({
 *      @ORM\AttributeOverride(name="totalInc",
 *          column=@ORM\Column(nullable= true, type="decimal", name="total_inc", scale=4, precision=20)
 *      ),
 *     @ORM\AttributeOverride(name="totalExcl",
 *          column=@ORM\Column(nullable= true, type="decimal", name="total_excl", scale=4, precision=20)
 *      ),
 *     @ORM\AttributeOverride(name="deliveryCost",
 *          column=@ORM\Column(nullable= true, type="decimal", name="delivery_cost", scale=4, precision=20)
 *      ),
 *     @ORM\AttributeOverride(name="deliveryVat",
 *          column=@ORM\Column(nullable= true, type="decimal", name="delivery_vat", scale=4, precision=20)
 *      )
 * })
 */
class Order extends BaseOrder
...
```
[...]

PR #485, opened 2017-12-03
Allow Symfony 3 and 4

ZRazor:
I'm targeting this branch because this PR breaks BC.

## Changelog

```markdown
### Fixed
- Allowed to install Symfony 3 and 4
### Changed
- Changed constructor arguments in `CustomerSelector`
- Replaced `ExecutionContextInterface` with `Context\ExecutionContextInterface`
- Replaced `ConstraintValidatorFactory` with `ContainerConstraintValidatorFactory `
```

## Subject
I think we should focus on master branch now, because adding support of SF 3 in 2.x with keeping BC will be too difficult.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,15 +36,17 @@
         "sonata-project/notification-bundle": "^3.3",
         "sonata-project/seo-bundle": "^2.4",
         "sonata-project/user-bundle": "^3.6 || ^4.0",
-        "symfony/config": "^2.8",
-        "symfony/console": "^2.8",
-        "symfony/form": "^2.8",
-        "symfony/http-foundation": "^2.8",
-        "symfony/process": "^2.8",
-        "symfony/routing": "^2.8",
-        "symfony/security-bundle": "^2.8",
-        "symfony/twig-bridge": "^2.8",
-        "symfony/validator": "^2.8"
+        "symfony/config": "^2.8 || ^3.2 || ^4.0",
+        "symfony/console": "^2.8 || ^3.2 || ^4.0",
+        "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
+        "symfony/form": "^2.8 || ^3.2 || ^4.0",
+        "symfony/http-foundation": "^2.8 || ^3.2 || ^4.0",
+        "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
+        "symfony/process": "^2.8 || ^3.2 || ^4.0",
+        "symfony/routing": "^2.8 || ^3.2 || ^4.0",
+        "symfony/security": "^2.8 || ^3.2 || ^4.0",
+        "symfony/twig-bridge": "^2.8 || ^3.2 || ^4.0",
+        "symfony/validator": "^2.8 || ^3.2 || ^4.0"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^3.3 || ^4.0"
--- a/src/Component/Customer/CustomerSelector.php
+++ b/src/Component/Customer/CustomerSelector.php
@@ -15,7 +15,8 @@ namespace Sonata\Component\Customer;
 
 use Sonata\IntlBundle\Locale\LocaleDetectorInterface;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
-use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
 
 class CustomerSelector implements CustomerSelectorInterface
@@ -23,34 +24,46 @@ class CustomerSelector implements CustomerSelectorInterface
     /**
      * @var \Sonata\Component\Customer\CustomerManagerInterface
      */
-    protected $customerManager;
+    private $customerManager;
 
     /**
      * @var \Symfony\Component\HttpFoundation\Session\Session
      */
-    protected $session;
+    private $session;
 
     /**
-     * @var SecurityContextInterface
+     * @var AuthorizationCheckerInterface
      */
-    protected $securityContext;
+    private $authorizationChecker;
+
+    /**
+     * @var TokenStorageInterface
+     */
+    private $tokenStorage;
 
     /**
      * @var string
      */
-    protected $locale;
+    private $locale;
 
     /**
-     * @param CustomerManagerInterface $customerManager
-     * @param SessionInterface         $session
-     * @param SecurityContextInterface $securityContext
-     * @param LocaleDetectorInterface  $localeDetector
+     * @param CustomerManagerInterface      $customerManager
+     * @param SessionInterface              $session
+     * @param AuthorizationCheckerInterface $authorizationChecker
+     * @param TokenStorageInterface         $tokenStorage
+     * @param LocaleDetectorInterface       $localeDetector
      */
-    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, SecurityContextInterface $securityContext, LocaleDetectorInterface $localeDetector)
-    {
+    public function __construct(
+        CustomerManagerInterface $customerManager,
+        SessionInterface $session,
+        AuthorizationCheckerInterface $authorizationChecker,
+        TokenStorageInterface $tokenStorage,
+        LocaleDetectorInterface $localeDetector
+    ) {
         $this->customerManager = $customerManager;
         $this->session = $session;
-        $this->securityContext = $securityContext;
+        $this->authorizationChecker = $authorizationChecker;
+        $this->tokenStorage = $tokenStorage;
         $this->locale = $localeDetector->getLocale();
     }
 
@@ -66,9 +79,9 @@ class CustomerSelector implements CustomerSelectorInterface
         $customer = null;
         $user = null;
 
-        if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_FULLY')) {
+        if ($this->authorizationChecker->isGranted('IS_AUTHENTICATED_FULLY')) {
             // user is authenticated
-            $user = $this->securityContext->getToken()->getUser();
+            $user = $this->tokenStorage->getToken()->getUser();
 
             if (!$user instanceof UserInterface) {
                 throw new \RuntimeException('User must be an instance of Symfony\Component\Security\Core\User\UserInterface');
--- a/src/Component/Product/ProductInterface.php
+++ b/src/Component/Product/ProductInterface.php
@@ -16,7 +16,7 @@ namespace Sonata\Component\Product;
 use Doctrine\Common\Collections\ArrayCollection;
 use Sonata\MediaBundle\Model\GalleryInterface;
 use Sonata\MediaBundle\Model\MediaInterface;
-use Symfony\Component\Validator\ExecutionContextInterface;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
 
 interface ProductInterface extends PriceComputableInterface
 {
--- a/tests/Component/Customer/CustomerSelectorTest.php
+++ b/tests/Component/Customer/CustomerSelectorTest.php
@@ -23,8 +23,9 @@ use Sonata\IntlBundle\Locale\LocaleDetectorInterface;
 use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\HttpFoundation\Session\Storage\MockArraySessionStorage;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
-use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
 class User
 {
@@ -47,13 +48,15 @@ class CustomerSelectorTest extends TestCase
 
         $session = $this->createMock(SessionInterface::class);
 
-        $securityContext = $this->createMock(SecurityContextInterface::class);
-        $securityContext->expects($this->once())->method('isGranted')->will($this->returnValue(false));
+        $authorizationChecker = $this->createMock(AuthorizationCheckerInterface::class);
+        $authorizationChecker->expects($this->once())->method('isGranted')->will($this->returnValue(false));
+
+        $tokenStorage = $this->createMock(TokenStorageInterface::class);
 
         $localeDetector = $this->createMock(LocaleDetectorInterface::class);
         $localeDetector->expects($this->once())->method('getLocale')->will($this->returnValue('en'));
 
-        $customerSelector = new CustomerSelector($customerManager, $session, $securityContext, $localeDetector);
+        $customerSelector = new CustomerSelector($customerManager, $session, $authorizationChecker, $tokenStorage, $localeDetector);
 
         $customer = $customerSelector->get();
 
@@ -72,14 +75,16 @@ class CustomerSelectorTest extends TestCase
         $token = $this->createMock(TokenInterface::class);
         $token->expects($this->once())->method('getUser')->will($this->returnValue(new User()));
 
-        $securityContext = $this->createMock(SecurityContextInterface::class);
-        $securityContext->expects($this->once())->method('isGranted')->will($this->returnValue(true));
-        $securityContext->expects($this->once())->method('getToken')->will($this->returnValue($token));
+        $authorizationChecker = $this->createMock(AuthorizationCheckerInterface::class);
+        $authorizationChecker->expects($this->once())->method('isGranted')->will($this->returnValue(true));
+
+        $tokenStorage = $this->createMock(TokenStorageInterface::class);
+        $tokenStorage->expects($this->once())->method('getToken')->will($this->returnValue($token));
 
         $localeDetector = $this->createMock(LocaleDetectorInterface::class);
         $localeDetector->expects($this->once())->method('getLocale')->will($this->returnValue('en'));
 
-        $customerSelector = new CustomerSelector($customerManager, $session, $securityContext, $localeDetector);
+        $customerSelector = new CustomerSelector($customerManager, $session, $authorizationChecker, $tokenStorage, $localeDetector);
 
         $customerSelector->get();
     }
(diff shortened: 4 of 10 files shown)


composer.json
```diff
@@ -12,40 +12,38 @@
...
+        "sonata-project/doctrine-orm-admin-bundle": "^3.2",
         "sonata-project/formatter-bundle": "^3.2",
-        "sonata-project/intl-bundle": "^2.3",
-        "sonata-project/media-bundle": "^3.7",
+        "sonata-project/intl-bundle": "^2.4",
+        "sonata-project/media-bundle": "^3.10",
         "sonata-project/notification-bundle": "^3.2",
-        "sonata-project/seo-bundle": "^2.2",
+        "sonata-project/seo-bundle": "^2.4",
         "sonata-project/user-bundle": "^3.5",
```
jordisala1991: This should be `^3.5 || ^4.0`

composer.json
```diff
@@ -12,40 +12,38 @@
     "require" : {
         "php": "^5.6 || ^7.0",
         "cocur/slugify": "^1.4 || ^2.5",
-        "friendsofsymfony/comment-bundle": "^2.0.13",
         "friendsofsymfony/rest-bundle": "^1.8 || ^2.2",
-        "friendsofsymfony/user-bundle": "^1.3",
```
jordisala1991: This bundle is used in the code, maybe `^1.3 || ^2.0`?

src/Component/Product/ProductInterface.php
```diff
@@ -14,7 +14,7 @@
 use Doctrine\Common\Collections\ArrayCollection;
 use Sonata\MediaBundle\Model\GalleryInterface;
 use Sonata\MediaBundle\Model\MediaInterface;
-use Symfony\Component\Validator\ExecutionContextInterface;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
```
jordisala1991: You cannot change this, maybe just remove the type hint on the function using it?

ZRazor: That will be incompatiable too. I'l remove this commit. Now I just need to fix travis build, because it installs symfony/validator 3.2 in one of the required builds.

jordisala1991: It is not BC break to remove a type hint, only to add it or change it from one to another

ZRazor: Thanks, I'll do it in a separate PR later.

composer.json
```diff
@@ -12,40 +12,39 @@
...
-        "symfony/console": "^2.8",
-        "symfony/form": "^2.8",
-        "symfony/http-foundation": "^2.8",
-        "symfony/process": "^2.8",
-        "symfony/routing": "^2.8",
-        "symfony/security-bundle": "^2.8",
-        "symfony/twig-bridge": "^2.8",
-        "symfony/validator": "^2.8"
+        "sonata-project/seo-bundle": "^2.4",
+        "sonata-project/user-bundle": "^3.5 || ^4.0",
```
jordisala1991: Probably until user-bundle 4.0 comes out you need to add ^4.0@dev

composer.json
ZRazor: Done

composer.json
jordisala1991: Not at this moment, maybe today will be

composer.json
```diff
@@ -12,40 +12,39 @@
     "require" : {
         "php": "^5.6 || ^7.0",
         "cocur/slugify": "^1.4 || ^2.5",
-        "friendsofsymfony/comment-bundle": "^2.0.13",
```
jordisala1991: Can we safely remove this one? Is it unused on the ecommerce?

ZRazor: It is already required here https://github.com/sonata-project/SonataCommentBundle/blob/3.x/composer.json#L21

composer.json
ZRazor: @jordisala1991 can you help?
I just tried to update composer with `"sonata-project/user-bundle": "^4.0@dev"` and `"sonata-project/user-bundle": "dev-master"` and got this:
```
    - sonata-project/user-bundle dev-master conflicts with sonata-project/intl-bundle[2.4.0].
    - sonata-project/intl-bundle 2.4.0 conflicts with sonata-project/user-bundle[dev-master].
    - don't install sonata-project/user-bundle 4.x-dev|install sonata-project/user-bundle dev-master
    - Installation request for sonata-project/user-bundle 4.x-dev -> satisfiable by sonata-project/user-bundle[4.x-dev].
    - Installation request for sonata-project/intl-bundle ^2.4 -> satisfiable by sonata-project/intl-bundle[2.4.0].
```
I checked: intl-bundle has
```
"conflict": {
        "sonata-project/user-bundle": "<2.0 || >=5.0"
    },
```
Why does composer throw this error?

jordisala1991: because intl-bundle depends on user-bundle and version 4.0 is not out yet, try ^3.0 || ^4.0@dev as 3.x-dev

composer.json
```diff
@@ -12,40 +12,38 @@
...
-        "symfony/console": "^2.8",
-        "symfony/form": "^2.8",
-        "symfony/http-foundation": "^2.8",
-        "symfony/process": "^2.8",
-        "symfony/routing": "^2.8",
-        "symfony/security-bundle": "^2.8",
-        "symfony/twig-bridge": "^2.8",
-        "symfony/validator": "^2.8"
+        "sonata-project/seo-bundle": "^2.4",
+        "sonata-project/user-bundle": "^4.0",
```
jordisala1991: You dont need to drop 3.0 AFAIK

composer.json
jordisala1991: It depends: Your code is using fosUser? then you habe to depend on it. Why? because dropping a dependency is not a BC break. We might drop FosUser from SonataUser and this bundle would break then. AFAIK we are using FosUser classes here

composer.json
ZRazor: I was just trying builds, of cause it should be ^3.5 || ^4.0

composer.json
jordisala1991: In this case I did a fast search and couldnt find code that uses this bundle directly

composer.json
ZRazor: It's only `FOS\UserBundle\Model\UserInterface` that is used in `CustomerInterface`, `CustomerSelector`, `BaseCustomer`.
I think, we should replace it with `Sonata\UserBundle\Model\UserInterface` and fully rely on SonataUserBundle.

jordisala1991: And why dont we use directly userinterface from symfony?

ZRazor: May be. I dont know why this specific check for `FOS\UserBundle\Model\UserInterface` was added.
`CustomerSelector`
```php
       if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_FULLY')) {
            // user is authenticated
            $user = $this->securityContext->getToken()->getUser();

            if (!$user instanceof UserInterface) {
                throw new \RuntimeException('User must be an instance of FOS\UserBundle\Model\UserInterface');
            }

            $customer = $this->customerManager->findOneBy([
                'user' => $user->getId(),
            ]);
        }
```

If we can safely replace it with `Symfony\Component\Security\Core\User\UserInterface` than we definitely should do it.

composer.json
```diff
@@ -12,40 +12,38 @@
...
-        "symfony/validator": "^2.8"
+        "sonata-project/seo-bundle": "^2.4",
+        "sonata-project/user-bundle": "^3.5 || ^4.0",
+        "symfony/config": "^2.8 || ^3.2 || ^4.0",
+        "symfony/console": "^2.8 || ^3.2 || ^4.0",
+        "symfony/form": "^2.8 || ^3.2 || ^4.0",
+        "symfony/http-foundation": "^2.8 || ^3.2 || ^4.0",
+        "symfony/process": "^2.8 || ^3.2 || ^4.0",
+        "symfony/routing": "^2.8 || ^3.2 || ^4.0",
+        "symfony/security-bundle": "^2.8 || ^3.2 || ^4.0",
```
jordisala1991: Maybe we can replace it by the component security instead?

composer.json
jordisala1991: Can you create a small PR with this replacement and drop user-bundle there?

ZRazor: Yes, but replace with `Sonata\UserBundle\Model\UserInterface` or `Symfony\Component\Security\Core\User\UserInterface` ?

I noticed that `Sonata\UserBundle\Model\UserInterface` is used in `BaseInvoice` and some other files. If someone is using checks like above, it will break BC.
So I think it should be `Sonata\UserBundle\Model\UserInterface`, at least for now.

jordisala1991: Changing a type hint to a parent is not a bc break. Thats why it is better to use symfony userinterface. Is the grandparent of all users

ZRazor: @jordisala1991 #490 check it, please

petitgen:
Great Job !!

composer.json
```diff
@@ -36,15 +36,15 @@
         "sonata-project/notification-bundle": "^3.3",
         "sonata-project/seo-bundle": "^2.4",
         "sonata-project/user-bundle": "^3.6 || ^4.0",
-        "symfony/config": "^2.8",
-        "symfony/console": "^2.8",
-        "symfony/form": "^2.8",
-        "symfony/http-foundation": "^2.8",
-        "symfony/process": "^2.8",
-        "symfony/routing": "^2.8",
-        "symfony/security-bundle": "^2.8",
```
jordisala1991: Did you check if we can move security-bundle to security directly?

Security bundle has this 3 dependencies.

        "symfony/security": "~3.4|~4.0",
        "symfony/dependency-injection": "~3.4|~4.0",
        "symfony/http-kernel": "~3.4|~4.0"

I did a fast search, and we need dependencyInjection and httpKernel too. Can you add them please?

jordisala1991:
I think there are only 2 things that needs to be fixed:

one is related to: 

```
1) Sonata\Component\Tests\Customer\CustomerSelectorTest::testUserNotConnected

Cannot stub or mock class or interface "Symfony\Component\Security\Core\SecurityContextInterface" which does not exist
```

and the other is related to: 

```
6) Sonata\Component\Tests\Form\BasketValidatorTest::testValidate

Cannot stub or mock class or interface "Symfony\Component\Validator\ExecutionContext" which does not exist
```

tests/Component/Customer/CustomerSelectorTest.php
```diff
@@ -150,7 +150,7 @@ public function testNonExistingCustomerInSession()
         $token = $this->createMock('Symfony\Component\Security\Core\Authentication\Token\TokenInterface');
         $token->expects($this->once())->method('getUser')->will($this->returnValue($user));
 
-        $securityContext = $this->createMock('Symfony\Component\Security\Core\SecurityContextInterface');
+        $securityContext = $this->createMock('Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface');
```
OskarStark: please use `::class` constant

tests/Component/Customer/CustomerSelectorTest.php
```diff
@@ -38,7 +38,7 @@ public function testUserNotConnected()
 
         $session = $this->createMock('Symfony\Component\HttpFoundation\Session\SessionInterface');
 
-        $securityContext = $this->createMock('Symfony\Component\Security\Core\SecurityContextInterface');
+        $securityContext = $this->createMock('Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface');
```
jordisala1991: use ::class notation instead:

```
use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;

AuthorizationCheckerInterface::class
```

You will also need to change the construct of the class, take a look at this: https://github.com/sonata-project/SonataAdminBundle/blob/4a3dbddff7b519e83a8e54f0fe51d7dcebd1d1f4/src/Security/Handler/RoleSecurityHandler.php#L37

We removed the type hint, and added a check for the type inside. Because you cannot change type hint without being BC break

jordisala1991: Also you might need to add another argument, because AuthorizationChecker does not have getToken. SecurityContext was split into 2 classes, I cannot remember the other one.

You might add an optional argument, and check for the 2 signatures.

ZRazor: I'll do it in a separate PR, thanks.

ZRazor: @jordisala1991 Am i going in the right direction?
New constructor:
```php
    /**
     * @param CustomerManagerInterface $customerManager
     * @param SessionInterface $session
     * @param $authorizationChecker
     * @param LocaleDetectorInterface $localeDetector
     * @param TokenStorageInterface $tokenStorage
     */
    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session,
                                $authorizationChecker, LocaleDetectorInterface $localeDetector,
                                TokenStorageInterface $tokenStorage = null)
    {
        $this->customerManager = $customerManager;
        $this->session = $session;
        // NEXT_MAJOR: Move AuthorizationCheckerInterface check to method signature.
        if (!$authorizationChecker instanceof AuthorizationCheckerInterface) {
            throw new \InvalidArgumentException('Argument 1 should be an instance of Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface');
        }
        $this->authorizationChecker = $authorizationChecker;
        $this->locale = $localeDetector->getLocale();
        $this->tokenStorage = $tokenStorage;
    }
```
And config will look (adding extra argument will not crash)
```xml
        <service id="sonata.customer.selector" class="%sonata.customer.selector.class%">
            <argument type="service" id="sonata.customer.manager"/>
            <argument type="service" id="session"/>
            <argument type="service" id="security.authorization_checker"/>
            <argument type="service" id="sonata.intl.locale_detector"/>
            <argument type="service" id="security.token_storage"/>
        </service>
```

jordisala1991: I think you do

ZRazor:
I changed target branch to `master` and updated PR.
@jordisala1991 @greg0ire what do you think about it?

.travis.yml
```diff
@@ -47,13 +47,19 @@ matrix:
...
+      env: SYMFONY=3.4.*
+    - php: '7.2'
+      env: SYMFONY='dev-master as 3.4.x-dev'
     - php: '7.2'
       env: SYMFONY_DEPRECATIONS_HELPER=0
   allow_failures:
     - php: nightly
     - env: SYMFONY_DEPRECATIONS_HELPER=0
-    - env: SYMFONY='dev-master as 2.8.x-dev'
+    - env: SYMFONY='dev-master as 3.4.x-dev'
```
greg0ire: This has to be changed through dev-kit

src/Component/Customer/CustomerSelector.php
```diff
@@ -31,26 +32,35 @@ class CustomerSelector implements CustomerSelectorInterface
...
+     * @param CustomerManagerInterface      $customerManager
+     * @param SessionInterface              $session
+     * @param AuthorizationCheckerInterface $authorizationChecker
+     * @param TokenStorageInterface         $tokenStorage
+     * @param LocaleDetectorInterface       $localeDetector
      */
-    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, SecurityContextInterface $securityContext, LocaleDetectorInterface $localeDetector)
+    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session,
+                                AuthorizationCheckerInterface $authorizationChecker, TokenStorageInterface $tokenStorage,
+                                LocaleDetectorInterface $localeDetector)
```
greg0ire: This is not PSR compliant, use one argument per line.

jordisala1991:
Why? this can be done in a BC way. Maybe this weekend I can help you finish this

src/Component/Customer/CustomerSelector.php
```diff
@@ -66,9 +76,9 @@ public function get()
         $customer = null;
         $user = null;
 
-        if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_FULLY')) {
+        if (true === $this->authorizationChecker->isGranted('IS_AUTHENTICATED_FULLY')) {
```
greg0ire: This method only returns boolean, no need to compare it with a boolean to get the final boolean

ZRazor:
@jordisala1991 It will be rather easly for users to upgrade to 3.0.
As for SF 3 - we need to
- Fix forms (remove SImpleChoiceList)
- Fix Api
- Make bundle compatiable with FOSUser 2 and SonataUser 4, beacase there are already things that wont work, like missing `BaseUserProfileBreadcrumbBlockService`. And only SonataUser 4 supports SF 3.
- Rename things like `AddressTypeType`
- Fix some bugs, that can also break BC (there are a few)
- Update docs
Also in master there is a fix for #404, which is quiet important.
This is why I think it's better to move to 3.x. By the way the most of other sonata bundles are already on 3.x and moving to 4.x and ecommerce is stuck at 2.x

jordisala1991:
Ok then, on master it is. We will most likely release a new major when this is completely done :)

ZRazor:
@jordisala1991 what else should I do to merge this one?)

src/Component/Customer/CustomerSelector.php
```diff
@@ -31,26 +32,37 @@ class CustomerSelector implements CustomerSelectorInterface
...
-     * @var SecurityContextInterface
+     * @var AuthorizationCheckerInterface
      */
-    protected $securityContext;
+    protected $authorizationChecker;
+
+    /**
+     * @var TokenStorageInterface
+     */
+    protected $tokenStorage;
```
greg0ire: Should be private.

src/Component/Customer/CustomerSelector.php
```diff
@@ -31,26 +32,37 @@ class CustomerSelector implements CustomerSelectorInterface
...
-     * @param SecurityContextInterface $securityContext
-     * @param LocaleDetectorInterface  $localeDetector
+     * @param CustomerManagerInterface      $customerManager
+     * @param SessionInterface              $session
+     * @param AuthorizationCheckerInterface $authorizationChecker
+     * @param TokenStorageInterface         $tokenStorage
+     * @param LocaleDetectorInterface       $localeDetector
      */
-    public function __construct(CustomerManagerInterface $customerManager, SessionInterface $session, SecurityContextInterface $securityContext, LocaleDetectorInterface $localeDetector)
+    public function __construct(CustomerManagerInterface $customerManager,
```
greg0ire: Please put the first and last argument on their own lines, this is not PSR2 compliant.

jordisala1991:
There is a build on travis on the allowed_failures section. You need to make it pass before merge ( it is testing Symfony dev-master with an alias to 2.8 )

src/Component/Customer/CustomerSelector.php
ZRazor: Yep, just today read PSRs and installed cs-fixer not to be wrong with this anymore )

ZRazor:
@jordisala1991 it's passing now.
@greg0ire I made all properties in `CustomerSelector` private (Made a note in `UPGRADE-3.0.md` too). The same thing should be done for other classes in future.

ZRazor:
So, what should I do next? 🌚

jordisala1991:
Thanks @ZRazor

PR #496, opened 2017-12-10
Fix Basket::removeBasketElement() for non existing products

ZRazor:
I am targeting this branch, because it's a bug fix.

## Changelog

```markdown
### Fixed
- Fixed clearing basket with deleted from db products
```

## Subject

When product exists in user's basket and is removed from db, it is automaticly removed from the basket (when user enters it) but it fails with error `Call to a member function getId() on null` in `removeBasketElement()` function.

`$element->getProduct()` returns null, and `$element->getProduct()->getId()` crashes.

### DIFF
--- a/tests/Component/Basket/BasketTest.php
+++ b/tests/Component/Basket/BasketTest.php
@@ -14,6 +14,7 @@ namespace Sonata\Component\Tests\Basket;
 use PHPUnit\Framework\TestCase;
 use Sonata\Component\Basket\Basket;
 use Sonata\Component\Basket\BasketElement;
+use Sonata\Component\Basket\BasketElementInterface;
 use Sonata\Component\Currency\CurrencyPriceCalculator;
 use Sonata\Component\Delivery\BaseServiceDelivery;
 use Sonata\Component\Product\Pool;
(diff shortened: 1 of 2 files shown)


greg0ire:
> When product exists in user's basket and is removed from db, it is automaticly removed form the basket (when user enters it) but it fails with error Call to a member function getId() on null in removeBasketElement() function.
> $element->getProduct() returns null, and $element->getProduct()->getId() crashes.

Please move this explanation to the commit message.

OskarStark:
Can you please add a testcase?

ZRazor:
@OskarStark, test added.
Checked it without fix - fail with `Call to a member function getId() on null`, with fix - passed.

tests/Component/Basket/BasketTest.php
```diff
@@ -413,6 +413,24 @@ public function testClean()
         $this->assertEquals(1, count($basket->getBasketElements()));
     }
 
+    public function testRemoveElementWithNotExistingProduct()
+    {
+        $basket = $this->getPreparedBasket();
+        $product = $this->getMockProduct();
+
+        $basketElement = $this->createMock('Sonata\Component\Basket\BasketElementInterface');
```
OskarStark: Please use `::class` constant wherever possible ❗️ 

Thank you 🙏

ZRazor: Yep, forgot it. We need a PR to replace these strings in all test files.

OskarStark: > We need a PR to replace these strings in all test files.

should be done through the whole code base and all sonata bundles 😄

greg0ire:
Thanks @ZRazor

PR #502, opened 2017-12-20
Fix ProductAdmin for SonataAdminBundle ^3.29

ZRazor:
I am targeting this branch, because it's a bug fix.

## Changelog

```markdown
### Fixed
- Fixed compatibility with SonataAdminBundle ^3.29
### Deprecated
- Deprecated `ProductAdmin::getProductClass` method
```

## Subject

After sonata-project/SonataAdminBundle#4824 `ProductAdmin::getClass`causes an infinite loop and is not needed anymore. I checked - on 3.29 it resolves different product providers correctly without overriding this method.
`ProductAdmin::getProductClass` method is also not needed anymore, so I deprecated it.

### DIFF
--- a/src/ProductBundle/Admin/ProductAdmin.php
+++ b/src/ProductBundle/Admin/ProductAdmin.php
@@ -61,32 +61,27 @@ class ProductAdmin extends AbstractAdmin
     /**
      * @return string
      */
-    public function getClass()
-    {
-        if ($this->hasSubject()) {
-            return get_class($this->getSubject());
-        } elseif ($class = $this->getProductClass()) {
-            return $class;
-        }
-
-        return parent::getClass();
-    }
-
-    /**
-     * @return string
-     */
     public function getProductType()
     {
         return $this->getRequest()->get('provider');
     }
 
     /**
+     * NEXT_MAJOR: remove this method.
+     *
      * Returns the product class from the provided request.
      *
      * @return string
+     *
+     * @deprecated since 2.2, will be removed in 3.0
      */
     public function getProductClass()
     {
+        @trigger_error(
+            'The '.__METHOD__.' method is deprecated since version 2.2 and will be removed in 3.0.',
+            E_USER_DEPRECATED
+        );
+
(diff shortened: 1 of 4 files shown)


greg0ire:
@Electricmaxx please review

src/ProductBundle/Admin/ProductAdmin.php
```diff
@@ -80,22 +66,6 @@ public function getProductType()
         return $this->getRequest()->get('provider');
     }
 
-    /**
-     * Returns the product class from the provided request.
-     *
-     * @return string
-     */
-    public function getProductClass()
```
jordisala1991: Not sure if we can allow this kind of BC break. Or we can allow it. The other way is just to deprecate the method

ZRazor: Sure, lets deprecate it.

src/ProductBundle/Admin/ProductAdmin.php
```diff
@@ -84,6 +70,8 @@ public function getProductType()
      * Returns the product class from the provided request.
      *
      * @return string
+     *
+     * @deprecated since 2.2, will be removed in 3.0
```
jordisala1991: It is not enough with this, you need also the trigger error

OskarStark: and please add a testcase which expects the deprecation

ZRazor: Done

ZRazor:
@OskarStark I had to bump `symfony/phpunit-bridge` from `^3.3 || ^4.0` to `^3.4 || ^4.0` to fix build on --prefer-lowest

ZRazor:
After we merge this, we should definitely make 2.2 release.

OskarStark:
> @OskarStark I had to bump symfony/phpunit-bridge from ^3.3 || ^4.0 to ^3.4 || ^4.0 to fix build on --prefer-lowest

IMO this is not a problem

PR #507, opened 2017-12-27
Fix php doc, fix types mismatches

ZRazor:
I am targeting this branch, because it's a BC break.

## Changelog

```markdown
### Fixed
- Fixed phpdoc
- Added missing params to `ProductManagerInterface` methods
- Added missing `parent::__construct` call in `BaseBasket`
- Replaced string types with FQCN in `BaseProductProvider`
- Fixed `Transformer/Pool` usage of indefined `methods` property
```

- Didn't touch Api controllers, they need a separate fix.

### DIFF
--- a/src/BasketBundle/DependencyInjection/SonataBasketExtension.php
+++ b/src/BasketBundle/DependencyInjection/SonataBasketExtension.php
@@ -65,8 +65,8 @@ class SonataBasketExtension extends Extension
     }
 
     /**
-     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
-     * @param $config
+     * @param ContainerBuilder $container
+     * @param array            $config
      */
     public function registerParameters(ContainerBuilder $container, array $config): void
     {
--- a/src/BasketBundle/Entity/BaseBasketElement.php
+++ b/src/BasketBundle/Entity/BaseBasketElement.php
@@ -19,14 +19,12 @@ use Sonata\Component\Basket\BasketInterface;
 abstract class BaseBasketElement extends BasketElement
 {
     /**
-     * @var \Sonata\Component\Basket\BasketInterface
+     * @var BasketInterface
      */
     protected $basket;
 
     /**
-     * Get basket.
-     *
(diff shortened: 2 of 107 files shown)


jordisala1991:
Can you explain which part is BC break? (it could be useful to isolate it maybe on a separate PR for master)

Another issue is with code comments, @greg0ire is working on comment removal for some bundles, we might need his help or look his PR to understand which comments we want to mantain and which not. The first one that comes to my mind is the inherit doc, he removed all of them from some bundles already.

Good work btw :)

src/BasketBundle/DependencyInjection/SonataBasketExtension.php
```diff
@@ -65,8 +65,8 @@ public function load(array $configs, ContainerBuilder $container): void
     }
 
     /**
-     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
-     * @param $config
+     * @param ContainerBuilder $container
+     * @param array            $config
      */
```
greg0ire: Can be removed

src/BasketBundle/Entity/BaseBasketElement.php
```diff
@@ -19,24 +19,20 @@
...
     public function getBasket()
     {
         return $this->basket;
     }
 
     /**
-     * Set basket.
-     *
-     * @param \Sonata\Component\Basket\BasketInterface $basket
+     * @param BasketInterface $basket
```
greg0ire: Can be removed

src/CustomerBundle/Entity/BaseAddress.php
```diff
@@ -233,18 +232,14 @@ public function setFirstname($firstname): void
     }
 
     /**
-     * Get firstname.
-     *
-     * @return string
+     * {@inheritdoc}
```
greg0ire: Can be removed

ZRazor:
I close this one and will make a new one with docs fix only for 2.x and separate fixes for master.

PR #517, opened 2018-02-12
Fixed dependency on SecurityContextInterface

Xymanek:
I am targeting this branch, because this is aimed at sf 3+.

## Changelog

```markdown
### Fixed
- Fixed missing security.context service error
```

### DIFF
--- a/src/OrderBundle/Block/RecentOrdersBlockService.php
+++ b/src/OrderBundle/Block/RecentOrdersBlockService.php
@@ -48,7 +48,7 @@ class RecentOrdersBlockService extends BaseBlockService
     protected $tokenStorage;
 
     /**
-     * @var Pool
+     * @var Pool|null
      */
     protected $adminPool;
 
@@ -58,7 +58,7 @@ class RecentOrdersBlockService extends BaseBlockService
         OrderManagerInterface $orderManager,
         CustomerManagerInterface $customerManager,
         TokenStorageInterface $tokenStorage,
-        ?Pool $adminPool = null
+        ?Pool $adminPool
     ) {
         $this->orderManager = $orderManager;
         $this->customerManager = $customerManager;
--- a/src/OrderBundle/Block/RecentOrdersBlockService.php
+++ b/src/OrderBundle/Block/RecentOrdersBlockService.php
@@ -24,7 +24,7 @@ use Sonata\CoreBundle\Validator\ErrorElement;
 use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
-use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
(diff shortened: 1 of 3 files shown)


kunicmarko20:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

src/OrderBundle/Block/RecentOrdersBlockService.php
```diff
@@ -73,7 +76,7 @@ public function execute(BlockContextInterface $blockContext, Response $response
         $criteria = [];
 
         if ('admin' !== $blockContext->getSetting('mode')) {
-            $orders = $this->orderManager->findForUser($this->securityContext->getToken()->getUser(), ['createdAt' => 'DESC'], $blockContext->getSetting('number'));
+            $orders = $this->orderManager->findForUser($this->tokenStorage->getToken()->getUser(), ['createdAt' => 'DESC'], $blockContext->getSetting('number'));
```
kunicmarko20: Could you break this line, we follow limit of 120 characters

Xymanek: I though about it, but running `php-cs-fixer` didn't change that....

Will fix now

src/OrderBundle/Resources/config/block.xml
```diff
@@ -7,8 +7,8 @@
             <argument type="service" id="templating"/>
             <argument type="service" id="sonata.order.manager"/>
             <argument type="service" id="sonata.customer.manager"/>
-            <argument type="service" id="security.context"/>
-            <argument type="service" id="sonata.admin.pool" on-invalid="ignore"/>
+            <argument type="service" id="security.token_storage"/>
+            <argument type="service" id="sonata.admin.pool" on-invalid="null"/>
```
jordisala1991: why the on-invalid?

Xymanek: No idea, it was there before

jordisala1991:
Can you review this one @greg0ire ? Do we need an upgrade note here?

greg0ire:
> Do we need an upgrade note here?

Yes. If it is major, then an action should be taken by the user after upgrading. This action should be described.

@Xymanek, please avoid "fix" in commit messages. Instead, describe precisely what you did and why you did it. More info in the saved reply below:

Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

Xymanek:
> If it is major, then an action should be taken by the user after upgrading

What action? I just used new sf security API instead of old one in an internal class - this should be transparent to users (developers)

greg0ire:
If it's transparent it is not major. But the signature changes, so it is a BC-break, and the action to take is to stop calling that class with the wrong argument if you do, and to change any class that inherits from this one if there are.

Xymanek:
> stop calling that class with the wrong argument if you do, and to change any class that inherits from this one if there are.

You are right, I didn't considered these cases... In that case does it also make sense to remove ` = null` from `$adminPool`?

greg0ire:
I think it does

src/OrderBundle/Block/RecentOrdersBlockService.php
```diff
@@ -43,23 +43,26 @@ class RecentOrdersBlockService extends BaseBlockService
...
-    {
+    protected $adminPool;
+
+    public function __construct(
+        string $name,
+        EngineInterface $templating,
+        OrderManagerInterface $orderManager,
+        CustomerManagerInterface $customerManager,
+        TokenStorageInterface $tokenStorage,
+        ?Pool $adminPool
```
jordisala1991: where is this used?

jordisala1991: oh I see it now

kunicmarko20:
```
return $this->renderPrivateResponse($blockContext->getTemplate(), [
             'context' => $blockContext,
             'settings' => $blockContext->getSettings(),
             'block' => $blockContext->getBlock(),
             'orders' => $orders,
             'admin_pool' => $this->adminPool,
         ], $response);
     }
```

src/OrderBundle/Resources/config/block.xml
jordisala1991: If you remove the null on the construct you need to remove the on-invalid here

kunicmarko20:
The constructor is nullable (?Pool), it just isn't null by default

jordisala1991:
Omg so many errors by me on this one.

Anyway I think that could be removed, it was added here: https://github.com/sonata-project/ecommerce/commit/cc8c1bb8831adb46bc6a483fe6f442d91da2235c

But the admin-bundle is a dependency on this bundle, so I guess the admin pool will be always be there... wdyt?

kunicmarko20:
Makes sense to remove it if we will always have the pool.

Xymanek:
Well it does check if pool is present or not in the template, so dunno...

https://github.com/Xymanek/ecommerce/blob/ef5af32faf5f560e437d3ee3d416f163395588e8/src/OrderBundle/Resources/views/Block/recent_orders.html.twig#L31

https://github.com/Xymanek/ecommerce/blob/ef5af32faf5f560e437d3ee3d416f163395588e8/src/OrderBundle/Resources/views/Block/recent_orders.html.twig#L53

jordisala1991:
Let's not remove the `on-invalid` then for now, it can be done later :P

jordisala1991:
Thank you @Xymanek

PR #522, opened 2018-02-13
Removed code that relies on missing features of UserBundle

Xymanek:
I am targeting this branch, because this is aimed at sf 3+ and userbundle 4.x

```markdown
## Changelog
### Removed
- `Sonata\OrderBundle\Block\Breadcrumb\UserOrderBreadcrumbBlockService`
- `Sonata\CustomerBundle\Block\Breadcrumb\CustomerAddressBreadcrumbBlockService`
- `sonata.customer.block.breadcrumb_address` service
- `sonata.order.block.breadcrumb_order` service
- `sonata.customer.block.breadcrumb_address.class` parameter
- `sonata.order.block.breadcrumb_order.class` parameter
```

Right now these files break sandbox since jms aop tries to load annotations from them. To be reversed when userbundle has these features implemented again.


jordisala1991:
I am not sure on removing this @greg0ire

kunicmarko20:
Do we need to remove this? @jordisala1991 said that he will revert the missing parts

greg0ire:
Ok, let's not remove it then.

Xymanek:
Currently this breaks `ecommerce` and sandbox as result. I think this should be removed for now and then reverted

jordisala1991:
If we don't declare those services on an xml file, it still breaks ecommerce? Maybe just commenting those xml definitions could be enough?

Xymanek:
Now that JMS AOP is gone, commenting them out will probably work

jordisala1991:
The only thing needed is the `BaseUserProfileBreadcrumbBlockService`?

Xymanek:
I think so

jordisala1991:
Let's do one thing, we close this one and start working on recovering code of UserBundle.

Here is the branch were I am working: https://github.com/jordisala1991/SonataUserBundle/tree/hotfix/recover-needed-classes

You can use it on the sandbox until it gets merged. For now it has the class you need

Xymanek:
👍

PR #523, opened 2018-02-17
Composer improvements

jordisala1991:
I am targeting this branch, because this is BC.

## Changelog

```markdown
### Added
- Allow Buzz `^0.16`
- Allow Cocur slugify `^3.0`

### Changed
- FOSRest, Nelmio Api Doc and JMS Serializer are no longer required dependencies

### Removed
- Compatibility with Cocur slugify `^1.0`
- Compatibility with Knp Menu `^1.0`
```

## Subject

- Added missing symfony dependencies
- FOSRest, Nelmio and JMS Serializer are always require-dev dependencies on Sonata bundles (I did check that the api is only loaded if those bundles are present)
- Removed security bundle in favor of the specific components we are using
- Up PHPUnit-bridge to 4.0


Xymanek:
> FOSRest, Nelmio Api Doc, [..] are no longer required dependencies

How come? Controllers use annotations provided by those bundles - they will fail to parse if those bundles are missing (and they will also not function)

jordisala1991:
What controllers? If you are thinking about api controllers, they are loaded only if FOSRest and nelmio are enabled, you can see that here: https://github.com/sonata-project/ecommerce/blob/2.x/src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php#L46

Xymanek:
Ah ok. maybe add them to `suggest` then?

jordisala1991:
Yes, it is a good idea. We are not using suggest section too much on sonata (and not correctly on some cases), but here we are starting fresh on suggest section.

Added like it is done here: https://github.com/sonata-project/SonataUserBundle/blob/4.x/composer.json#L66

greg0ire:
Thanks @jordisala1991 !

PR #525, opened 2018-02-24
Wrong argument type

kunicmarko20:
I am targeting this branch, because this is BC fix.

```markdown
### Fixed
- wrong argument types that were passed to multiple functions
```
## Subject

Not sure if this ever worked as this functions expect string, so here is a fix.

### DIFF
--- a/src/Component/Basket/Basket.php
+++ b/src/Component/Basket/Basket.php
@@ -429,7 +429,7 @@ class Basket implements \Serializable, BasketInterface
      */
     public function getTotal($vat = false, $recurrentOnly = null)
     {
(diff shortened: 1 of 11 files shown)


jordisala1991:
Maybe you can look at this other commit: https://github.com/sonata-project/ecommerce/pull/497/commits/d5c6b4e04a3326bf17897f3a9490a54787eb4e11

I had to fix this on master to make test pass. At that moment we didn't backport it to 2.x, maybe it is time :P

kunicmarko20:
oh, there is more. Will try to port it to 2.x then.

PR #528, opened 2018-02-25
Manual merge

kunicmarko20:
Just merge, don't approve.

### DIFF
--- a/src/Component/Basket/BasketElement.php
+++ b/src/Component/Basket/BasketElement.php
@@ -277,14 +277,14 @@ class BasketElement implements \Serializable, BasketElementInterface
      */
     public function getPrice($vat = false)
     {
-        $price = $this->price;
+        $price = (string) $this->price;
 
         if (!$vat && true === $this->isPriceIncludingVat()) {
-            $price = bcdiv($price, bcadd(1, bcdiv($this->getVatRate(), 100)));
+            $price = bcdiv($price, bcadd('1', bcdiv((string) $this->getVatRate(), '100')));
         }
 
         if ($vat && false === $this->isPriceIncludingVat()) {
-            $price = bcmul($price, bcadd(1, bcdiv($this->getVatRate(), 100)));
+            $price = bcmul($price, bcadd('1', bcdiv((string) $this->getVatRate(), '100')));
         }
 
         return $price;
--- a/composer.json
+++ b/composer.json
@@ -57,8 +57,7 @@
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.2 || >=3.0",
-        "jms/serializer": "<0.13",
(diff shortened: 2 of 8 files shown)


src/Component/Basket/BasketElement.php
```diff
@@ -202,11 +202,11 @@ public function getUnitPrice($vat = false)
         $price = (string) $this->unitPrice;
 
         if (!$vat && $this->isPriceIncludingVat()) {
-            $price = bcdiv($price, bcadd('1', bcdiv((string) $this->getVatRate(), '100')));
+            $price = bcdiv($price, bcadd('1', (string) bcdiv((string) $this->getVatRate(), '100')));
```
jordisala1991: some of this is unnecesary, please have a look at the whole type hint thing, the original code already has some (string) because this change was already on master

kunicmarko20: bcdiv can return null `http://php.net/manual/en/function.bcdiv.php#refsect1-function.bcdiv-returnvalues` so this is needed because bcadd expects string?

kunicmarko20: ah, but our divisior is not 0, my bad, will fix it

kunicmarko20: fixed.

src/Component/Basket/BasketElement.php
```diff
@@ -206,7 +206,7 @@ public function getUnitPrice($vat = false)
         }
 
         if ($vat && !$this->isPriceIncludingVat()) {
-            $price = bcmul($price, bcadd('1', bcdiv((string) $this->getVatRate(), '100')));
+            $price = bcmul((string) $price, bcadd('1', bcdiv((string) $this->getVatRate(), '100')));
```
jordisala1991: $price is already a string, see above

src/Component/Basket/BasketElement.php
```diff
@@ -280,11 +280,11 @@ public function getPrice($vat = false)
         $price = $this->price;
 
         if (!$vat && true === $this->isPriceIncludingVat()) {
-            $price = bcdiv($price, bcadd(1, bcdiv($this->getVatRate(), 100)));
+            $price = bcdiv($price, bcadd('1', bcdiv($this->getVatRate(), '100')));
```
jordisala1991: this $price used in bcdiv needs a cast

jordisala1991:
Not sure if my review changes makes sense, it was wrong on 2.x or got wrong when merging?

kunicmarko20:
They make sense, but I think you did not catch everything with that pr on tests, I needed to fix someone additional things to get fixtures work on the sandbox and now it is merging to master. I fixed the things you pointed out.

PR #530, opened 2018-03-09
Cast arguments to right type

kunicmarko20:
I am targeting this branch, because this is BC fix.

## Subject

I guess I missed few types last time, hopefully, there is no more.


greg0ire:
Does this really fix something on the stable branch? Or does it fix something on the master branch after being merged into it. If it's the latter, I think this should be labeled pedantic, because nobody awaits a release for this, which means you could remove the changelog.

kunicmarko20:
Changed.

jordisala1991:
Thank you @kunicmarko20

PR #542, opened 2018-07-06
Clean PHPDoc

core23:
I am targeting this branch, because this is BC.

## Subject

Fixed some PHPDoc

### DIFF
--- a/src/ProductBundle/Command/GenerateProductCommand.php
+++ b/src/ProductBundle/Command/GenerateProductCommand.php
@@ -45,8 +45,10 @@ class GenerateProductCommand extends ContainerAwareCommand
     /**
      * @see Command
      *
-     * @throws \InvalidArgumentException When namespace doesn't end with Bundle
-     * @throws \RuntimeException         When bundle can't be executed
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     *
+     * @throws \Exception
      */
     protected function execute(InputInterface $input, OutputInterface $output): void
     {
--- a/src/ProductBundle/Command/GenerateProductCommand.php
+++ b/src/ProductBundle/Command/GenerateProductCommand.php
@@ -27,9 +27,6 @@ use Symfony\Component\Filesystem\Filesystem;
  */
 class GenerateProductCommand extends ContainerAwareCommand
 {
-    /**
-     * @see Command
-     */
     protected function configure(): void
     {
         $this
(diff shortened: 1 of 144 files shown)


src/ProductBundle/Model/BaseProductProvider.php
```diff
@@ -767,9 +704,12 @@ public function synchronizeVariationsPackages(ProductInterface $product, ArrayCo
     // BASKET RELATED FUNCTIONS
 
     /**
-     * (non-PHPdoc).
-     *
      * @see \Sonata\Component\Product\ProductProviderInterface::createBasketElement()
+     *
+     * @param ProductInterface|null $product
+     * @param array                 $options
```
kunicmarko20: these types are already defined,

src/ProductBundle/DependencyInjection/SonataProductExtension.php
```diff
@@ -28,8 +28,10 @@ class SonataProductExtension extends Extension
     /**
      * Loads the product configuration.
      *
-     * @param array            $configs   An array of configuration settings
+     * @param array            $configs An array of configuration settings
      * @param ContainerBuilder $container A ContainerBuilder instance
```
kunicmarko20: types are already defined, don't see any value in comments as the names explain well

src/ProductBundle/Controller/ProductVariationAdminController.php
```diff
@@ -23,10 +23,11 @@
 class ProductVariationAdminController extends Controller
 {
     /**
-     * @throws \Symfony\Component\Security\Core\Exception\AccessDeniedException
-     * @throws \Symfony\Component\HttpKernel\Exception\NotFoundHttpException
+     * @param Request|null $request
```
kunicmarko20: not needed

src/ProductBundle/Command/GenerateProductCommand.php
```diff
@@ -45,8 +45,10 @@ protected function configure(): void
     /**
      * @see Command
      *
-     * @throws \InvalidArgumentException When namespace doesn't end with Bundle
-     * @throws \RuntimeException         When bundle can't be executed
+     * @param InputInterface  $input
+     * @param OutputInterface $output
```
kunicmarko20: not needed

kunicmarko20:
We have 7.1 on master, half of this param comments are not needed, also return type comments could be replaced with a return type.

PR #543, opened 2018-07-06
Fixed GenerateProductCommand

core23:
I am targeting this branch, because the 2.x branch looks dead.

## Changelog

```markdown
### Fixed
- Add quotes to yaml strings when generating it
- Generated product classes start with an uppercase char
- Fixed wrong argument for `*ProductProvider`
```

## Subject

This are some fixes for the product generator.

### DIFF
--- a/src/ProductBundle/Command/GenerateProductCommand.php
+++ b/src/ProductBundle/Command/GenerateProductCommand.php
@@ -149,7 +149,7 @@ class GenerateProductCommand extends ContainerAwareCommand
     {{ service }}.type:
         class: Application\Sonata\ProductBundle\Provider\{{ product }}ProductProvider
         arguments:
-            - "@serializer"
+            - "@jms_serializer"
 
 <info>2. Add this service configuration</info>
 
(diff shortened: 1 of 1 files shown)


src/ProductBundle/Command/GenerateProductCommand.php
```diff
@@ -131,7 +131,7 @@ protected function execute(InputInterface $input, OutputInterface $output): void
             sprintf('%s/Resources/views/Entity', $bundle_dir),
         ]);
 
-        $product = $input->getArgument('product');
+        $product = ucfirst($input->getArgument('product'));
```
covex-nn: Is this the only place, where input argument must be modified? I found 7 places in `execute`, where `$input->getArgument('product')` is called

kunicmarko20:
Does it make sense to tag this with patch as it is master?

core23:
> Does it make sense to tag this with patch as it is master?

In general, yes. But there is no activity on the *stable* branch. I will try to make the master the new stable branch.

kunicmarko20:
I see, great work!

PR #551, opened 2019-01-02
Change authentication check to REMEMBERED instead of FULLY in customerSelector

### LINKED ISSUE #549: user not set in customer entity and multiple customer entities are created
## User reference missing in Customer record
I don't have much to go on but for some reason the user is not always set in the customer record in the database. 

Our application is fully behind a login, so to interact with any e-commerce related function the user must be logged in. 

When a visitor manages to create a customer without a user attached, he also creates multiple customer records (this makes sense because we can't link the customer to the logged in user).

## Steps to reproduce
The problem is that we can't reproduce this issue ourselves. So my hope is on someone recognizing this problem or pointing me in the right direction. 

## Expected results
User is always set in the Customer entity because the whole application is behind a login form. 

## Actual results
User not set in some cases and multiple customer entities get created.

jeroendk:
I am targeting this branch, because this is a bugfix that is backwards compatible.

Closes #549

## Changelog
- Fix CustomerSelector, users authenticated by the "remember me" cookie will now be linked to the created customer. 

```markdown

### Fixed
Changed the authentication check in the CustomerSelector, check for IS_AUTHENTICATED_REMEMBERED.
```

### DIFF
--- a/src/Component/Customer/CustomerSelector.php
+++ b/src/Component/Customer/CustomerSelector.php
@@ -66,7 +66,7 @@ class CustomerSelector implements CustomerSelectorInterface
         $customer = null;
         $user = null;
 
-        if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_FULLY')) {
+        if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')) {
             // user is authenticated
             $user = $this->securityContext->getToken()->getUser();
 


greg0ire:
Please fix the Travis and FlintCI

greg0ire:
Also, please rebase and use `git push --force` to get rid of the merge commit

jeroendk:
> Also, please rebase and use `git push --force` to get rid of the merge commit

Done!

> Please fix the Travis and FlintCI

Should I fix these issues even though they are not related to the file I changed? Code stlye issues in non related files and tests are failing in different components. 

Also the line that I changed is diffrent in the master branch compared to the 2.x branch:

Master
``if ($this->authorizationChecker->isGranted('IS_AUTHENTICATED_FULLY')) {``
2.x
``if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')) {``

Should I do something with this or will this be handled in a merge?

greg0ire:
We cannot merge unless Travis is green, it is a required check. This is a separate issue, but it blocks your PR (and maybe others).

jeroendk:
Well, I can try and fix the issues but not sure if I have enough knowledge of all the ecommerce bundles. Should this be done in a separate Pull Request?

greg0ire:
> Should this be done in a separate Pull Request

It would be better, but you're not forced to, it's up to you.

jeroendk:
@greg0ire This one can move forward now right?

src/Component/Customer/CustomerSelector.php
```diff
@@ -66,7 +66,7 @@ public function get()
         $customer = null;
         $user = null;
 
-        if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_FULLY')) {
+        if (true === $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')) {
```
kunicmarko20: shouldn't this check both? or if you are `fully` you are also `remembered`?

jeroendk: If I understand it correctly you are also `remembered` if you are `fully`:
https://symfony.com/doc/current/security/remember_me.html#forcing-the-user-to-re-authenticate-before-accessing-certain-resources

```
// allow any authenticated user - we don't care if they just
// logged in, or are logged in via a remember me cookie
$this->denyAccessUnlessGranted('IS_AUTHENTICATED_REMEMBERED');
```

greg0ire:
Thanks @jeroendk !

PR #555, opened 2019-01-13
Applied fixes from FlintCI

Soullivaneuh:


src/Component/Payment/BasePayment.php
```diff
@@ -66,7 +66,9 @@ public function generateUrlCheck(OrderInterface $order)
             throw new \RuntimeException(sprintf('The order must have a creation date - id:%s, reference:%s ', $order->getId(), $order->getReference()));
         }
 
-        var_dump($order->getCreatedAt()->format('m/d/Y:G:i:s'));die();
+        var_dump($order->getCreatedAt()->format('m/d/Y:G:i:s'));
+        die();
```
OskarStark: 🤔 😄

greg0ire: Can't even debug without FlintCI having something to say about it :P

PR #565, opened 2019-02-12
Make services public

pierredup:
## Subject

In Symfony 3.4, services are private by default, which means that it can't be retrieved from the dumped container using the `get` method unless they are explicitly set to public.
These are all the services that I could find that are pulled directly from the container, which breaks in Symfony 4 (in 3.4 it will still work due to the BC layer)

I am targeting this branch, because it adds support for Symfony 3 and 4.

## Changelog

```markdown
### Fixed
- service definitions to be public
```

### DIFF
--- a/src/BasketBundle/Resources/config/basket.xml
+++ b/src/BasketBundle/Resources/config/basket.xml
@@ -10,7 +10,7 @@
             <argument type="service" id="sonata.delivery.pool"/>
             <argument type="service" id="sonata.payment.pool"/>
         </service>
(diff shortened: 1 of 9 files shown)


pierredup:
> stop pulling services directly from the container

Yeah that would be the next step, but it's a fair amount of work, so I first wanted to get rid of the immediate failures

OskarStark:
Great work so far for Sonata-Project 👍🏻
Thank you @pierredup 🙏

PR #566, opened 2019-02-12
 Get product class from provider

### LINKED ISSUE #511: Product class in not detecting correctly while creating a new product
## Subject

In #502 I removed `ProductAdmin::getClass` method to avoid recursion after sonata-admin-bundle update. There is no more recursion in product editing, but in product creating class is not detected correctly (my bad 🙃 ).
I want to fix this, but I need some help.
We need to detect product class manualy (using `provider` parameter from query) only in create action. Before this was done by `if ($this->hasSubject())` inside `getClass` method, but we cannot do this now, because it stucks in recursion (since `hasSubject` calls `getClass`). 
@greg0ire @jordisala1991 Do you have any ideas how to do it right? 
I thought about `if ($this->getRequest()->get('_route') == 'admin_sonata_product_product_create')` but it's obviously not a good idea.

pierredup:
## Subject

When creating a new product, the product class is not correctly retrieved since the `provider` query parameter is not taken into account.

I am targeting this branch, because the issue was first introduced in the master branch in #502 

Closes #511

## Changelog

```markdown
### Fixed
- get product class from provider parameter
```

### DIFF
--- a/src/ProductBundle/Admin/ProductAdmin.php
+++ b/src/ProductBundle/Admin/ProductAdmin.php
@@ -166,6 +166,15 @@ class ProductAdmin extends AbstractAdmin
         return $product;
     }
 
+    public function getClass()
+    {
+        if ($this->hasRequest() && $code = $this->getProductType()) {
+            return $this->getProductPool()->getManager($code)->getClass();
+        }
+
+        return parent::getClass();
+    }
+
     public function validate(ErrorElement $errorElement, $object): void
     {
         $errorElement


greg0ire:
We have cs issues that were just fixed, I force pushed your branch, hopefully that will fix it.

greg0ire:
@ZRazor please review this

ZRazor:
There were several issues with this method, but I don't remember the details ;(

greg0ire:
Great! Thank you all!

PR #567, opened 2019-02-15
Remove deprecated Breadcrumb blocks

pierredup:
## Subject

The `BaseUserProfileBreadcrumbBlockService` block has been removed in SonataUserBundle version 4. This proposal is to remove support for SonataUserBundle 3 (Only supporting version 4) and remove the services that rely on non-existent classes.

Note: The stable version currently also supports version 4 of SonataUserBundle and registers these services

I am targeting this branch, because this might potentially be a BC break if someone relied on these services or classes to exist.

## Changelog

```markdown
### Removed
- the `sonata.customer.block.breadcrumb_address` service
- the `sonata.order.block.breadcrumb_order` service
```


greg0ire:
Thanks @pierredup !

PR #550, opened 2018-12-20
Fix 4.0

quisse:
I am targeting this branch, because not backwards compatible.

Closes #548 #531 #470 #511 

### Changed
- Allow SF4 formatter

## To do

- [ ] Update the tests
- [x] Update the documentation
- [ ] Add an upgrade note

## Subject

Allow SF4 formatter so this bundle works with SF4

PR #568, opened 2019-03-03
Add dedicated method to get templates path for product

pierredup:
## Subject

Since Symfony 4, Bundle inheritance has been removed. This means using a namespace like `SonataProductBundle:MyProduct:template.html.twig` doesn't work anymore, since it won't pick up the template in the child bundle anymore. The same goes for specifying a custom controller.
This makes it difficult to use a single namespace referencing the templates path as well as a controller name. To keep to the suggested best practices of Symfony, this PR adds a new `getTemplatesPath` method to the product provider, which can be used to specify the directory to load product templates from. This method defaults to the namespace of the child bundle, but can easily be changed to reference the top-level `templates` directory. Controllers are also now referenced using their FQCN

I am targeting this branch, because it's the only branch that supports Symfony 3/4.

## Changelog

```markdown
### Added
- Added `getTemplatesPath` method to product providers to specify path for product templates

```


src/ProductBundle/Resources/skeleton/product/Provider/EntityProductProvider.php
```diff
@@ -11,6 +11,7 @@
 namespace Application\Sonata\ProductBundle\Provider;
 
 use Sonata\ProductBundle\Model\BaseProductProvider;
+use Application\Sonata\ProductBundle\Controller\{{ product }}Controller;
```
pierredup: Once #550 is merged, then this can change to `App\Application\...`

OskarStark:
Can you please check Travis?

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i 3.x@{upstream}`
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already pushed. Don't forget the `--force` option otherwise git will try to merge both things together.

src/Component/Product/ProductProviderInterface.php
```diff
@@ -61,6 +61,11 @@ public function getProductCollectionManager();
      */
     public function getBaseControllerName();
 
+    /**
+     * @return string
+     */
+    public function getTemplatesPath();
```
greg0ire: Please remove the docblock above or better, add a description, and use a return type hint.
```suggestion
    public function getTemplatesPath(): string;
```

src/ProductBundle/Controller/BaseProductController.php
```diff
@@ -89,9 +90,11 @@ public function renderPropertiesAction(ProductInterface $product)
      */
     public function renderFormBasketElementAction(FormView $formView, BasketElementInterface $basketElement, BasketInterface $basket)
     {
-        $provider = $this->get('sonata.product.pool')->getProvider($basketElement->getProduct());
+        /** @var Pool $pool */
+        $pool = $this->get('sonata.product.pool');
```
greg0ire: Please add a constructor with one optional argument. If the argument is provided, use it, and if not, please trigger a deprecation error and default to this.

pierredup: There are various other places in this controller that this service is used, and I don't want to make too many unrelated changes to this PR. So should only this instance be updated, or can this stay as-is for now then everything can be updated in on go in a separate PR?

greg0ire: It can stay as is, I didn't know about the other occurences 👍

src/Component/Product/ProductProviderInterface.php
```diff
@@ -61,6 +61,11 @@ public function getProductCollectionManager();
      */
     public function getBaseControllerName();
 
+    /**
+     * @return string
+     */
```
greg0ire: Please remove this dockblock, it is tautological now.

src/ProductBundle/Resources/skeleton/product/Provider/EntityProductProvider.php
```diff
@@ -26,6 +27,11 @@ class {{ product }}ProductProvider extends BaseProductProvider
 {
     public function getBaseControllerName()
     {
-        return 'SonataProductBundle:{{ product }}';
+        return {{ product }}Controller::class;
+    }
+
+    public function getTemplatesPath()
```
greg0ire: You will have to add the return type hint here and below
```suggestion
    public function getTemplatesPath(): string
```

tests/Component/Basket/BasketElementTest.php
```diff
@@ -35,6 +35,14 @@ public function getBaseControllerName()
     {
         return 'DumbTestController';
     }
+
+    /**
+     * @return string
+     */
+    public function getTemplatesPath()
```
greg0ire:
```suggestion
    public function getTemplatesPath(): string
```

tests/Component/Product/BaseProductServiceTest.php
```diff
@@ -127,7 +127,15 @@ class BaseProductServiceTest_ProductProvider extends BaseProductProvider
...
     public function getBaseControllerName()
     {
-        // TODO: Implement getBaseControllerName() method.
+        return 'DumbTestController';
+    }
+
+    /**
+     * @return string
+     */
+    public function getTemplatesPath()
```
greg0ire:
```suggestion
    public function getTemplatesPath(): string
```

tests/ProductBundle/Model/BaseProductProviderTest.php
```diff
@@ -38,6 +38,14 @@ public function getBaseControllerName()
     {
         return 'DumbTestController';
     }
+
+    /**
+     * @return string
+     */
+    public function getTemplatesPath()
```
greg0ire:
```suggestion
    public function getTemplatesPath(): string
```

src/Component/Product/ProductProviderInterface.php
greg0ire: (here and below)

greg0ire:
Thanks @pierredup !

PR #569, opened 2019-03-03
Change template reference to use recommended native Twig reference

pierredup:
## Subject

Change all templates reference to use the recommended native Twig reference

`@SonataProductBundle/Path/template.html.twig` instead of `SonataProductBundle:Path:template.html.twig`

I am targeting this branch, because it is the recommended way to reference and override template since Symfony 3.


greg0ire:
This is a great contribution, but I don't think the end-user has a way to notice it besides having a look at the code. If that is the case, please remove the changelog and we will label this as pedantic.

pierredup:
@greg0ire 👍 Done

PR #562, opened 2019-02-10
Update forms compatibility with Symfony 3

pierredup:
## Subject

Update all forms to be compatible with Symfony 3.4 

I am targeting this branch, because it is the only branch currently supporting Symfony 3.

One step closer to Close #531

## Changelog

```markdown
### Changed
- form submissions to check `isSubmitted` before `isValid`
- form factory to use FQCN form FormType instead of `form` alias

### Removed
- `getName` method on all form types
```

PR #570, opened 2019-03-08
Fix forms to be compatible with Symfony 3.0+

pierredup:
## Subject

Fix forms to be compatible with Symfony 3+. These are a few instances that I missed in #562

I am targeting this branch, because it is the only branch currently supporting Symfony 3.

## Changelog

I'm not sure if this requires a changelog entry, since it's just a follow-up of a previous PR with similar changes?

### DIFF
--- a/src/Component/Form/Type/DeliveryChoiceType.php
+++ b/src/Component/Form/Type/DeliveryChoiceType.php
@@ -16,7 +16,7 @@ namespace Sonata\Component\Form\Type;
 use Sonata\Component\Delivery\Pool;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
-use Symfony\Component\OptionsResolver\OptionsResolverInterface;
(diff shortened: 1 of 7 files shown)


src/Component/Form/Type/DeliveryChoiceType.php
```diff
@@ -40,12 +40,7 @@ public function getBlockPrefix()
         return 'sonata_delivery_choice';
     }
 
-    public function getName()
-    {
-        return $this->getBlockPrefix();
-    }
-
-    public function setDefaultOptions(OptionsResolverInterface $resolver): void
+    public function ConfigureOptions(OptionsResolver $resolver): void
```
OskarStark: C -> c

greg0ire:
> I'm not sure if this requires a changelog entry, since it's just a follow-up of a previous PR with similar changes?

No need, since your changes were not released yet.

greg0ire:
Thanks @pierredup !

PR #571, opened 2019-03-08
Fix deprecations

pierredup:
## Subject

Sonata CoreBundle includes various deprecations in version 3 which will be removed in version 4.
This removes the usage of all the deprecated classes and interfaces for forward-compatibility.

I am targeting this branch, because it's bugfix against the current stable version.

## Changelog

```markdown
### Fixed
- removed usage of deprecated classes and interfaces
```

### DIFF
--- a/src/BasketBundle/Controller/Api/BasketController.php
+++ b/src/BasketBundle/Controller/Api/BasketController.php
@@ -563,7 +563,7 @@ class BasketController
     {
         $basket = $this->basketManager->findOneBy(['customer' => $customerId]);
         if ($basket instanceof BasketInterface) {
-            throw new HttpException('400', sprintf('Customer (%d) already has a basket', $customerId));
+            throw new HttpException(400, sprintf('Customer (%d) already has a basket', $customerId));
         }
     }
 }
--- a/src/CustomerBundle/Twig/Extension/AddressExtension.php
+++ b/src/CustomerBundle/Twig/Extension/AddressExtension.php
@@ -16,7 +16,6 @@ namespace Sonata\CustomerBundle\Twig\Extension;
 use Sonata\Component\Basket\BasketInterface;
 use Sonata\Component\Customer\AddressInterface;
 use Sonata\Component\Delivery\ServiceDeliverySelectorInterface;
-use Sonata\CoreBundle\Exception\InvalidParameterException;
 use Sonata\CustomerBundle\Entity\BaseAddress;
 
 /**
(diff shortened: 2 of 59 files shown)


src/BasketBundle/Controller/Api/BasketController.php
```diff
@@ -563,7 +563,7 @@ protected function checkExistingCustomerBasket($customerId)
     {
         $basket = $this->basketManager->findOneBy(['customer' => $customerId]);
         if ($basket instanceof BasketInterface) {
-            throw new HttpException('400', sprintf('Customer (%d) already has a basket', $customerId));
+            throw new HttpException(400, sprintf('Customer (%d) already has a basket', $customerId));
```
greg0ire: Please move this to a separate commit, it does not have to do with the core bundle, same for any unrelated change below.

greg0ire: I did it for you, and actually made 4 commits.

src/CustomerBundle/Twig/Extension/AddressExtension.php
```diff
@@ -81,7 +80,7 @@ public function renderAddress(\Twig_Environment $environment, $address, $showNam
         $requiredAddressKeys = ['firstname', 'lastname', 'address1', 'postcode', 'city', 'country_code'];
 
         if (!($address instanceof AddressInterface) && (!\is_array($address) || 0 !== \count(array_diff($requiredAddressKeys, array_keys($address))))) {
-            throw new InvalidParameterException(sprintf(
+            throw new \RuntimeException(sprintf(
```
greg0ire: @core23 , can you confirm `InvalidParameterException` is deprecated in favor of `\RuntimeException`?

greg0ire:
After this one I will do a release because the last one is so old :sweat_smile:

src/CustomerBundle/Twig/Extension/AddressExtension.php
core23: Yes https://github.com/sonata-project/SonataCoreBundle/blob/3.x/src/CoreBundle/Exception/InvalidParameterException.php#L16-L18

core23:
> After this one I will do a release because the last one is so old 😅

Have you tested this project? Not sure if this project is working with symfony 3/4...

greg0ire:
I mean a minor release, not a major one, no need for any tests for that, right?

src/CustomerBundle/Twig/Extension/AddressExtension.php
greg0ire: It doesn't say in favor of what it is deprecated, but ok!

PR #572, opened 2019-03-10
Replace render tag with function in Twig templates

pierredup:
## Subject

The `{% render %}` tag has been removed in Symfony 3 and replaced with the `{{ render() }}` function

I am targeting this branch, because it's the only branch supporting Symfony 3.

## Changelog

```markdown
### fixed
- replaced the render tag in twig with the render function
```


greg0ire:
Thanks @pierredup !

PR #574, opened 2019-03-10
bump formatter version

quisse:
## Subject

I am targeting this branch, because not backwards compatible.

## Changelog

```markdown
### Changed
- newer formatter bundle version
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "sonata-project/core-bundle": "^3.13.0",
         "sonata-project/datagrid-bundle": "^2.3",
         "sonata-project/doctrine-orm-admin-bundle": "^3.2",
-        "sonata-project/formatter-bundle": "^3.4",
+        "sonata-project/formatter-bundle": "^3.4 || ^4.0",
         "sonata-project/intl-bundle": "^2.4",
         "sonata-project/media-bundle": "^3.10",
         "sonata-project/notification-bundle": "^3.3",
--- a/tests/ProductBundle/Controller/Api/ProductControllerTest.php
+++ b/tests/ProductBundle/Controller/Api/ProductControllerTest.php
@@ -16,6 +16,7 @@ namespace Sonata\ProductBundle\Tests\Controller\Api;
 use FOS\RestBundle\Request\ParamFetcherInterface;
 use FOS\RestBundle\View\View;
 use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
 use Sonata\ClassificationBundle\Model\CategoryInterface;
 use Sonata\ClassificationBundle\Model\CollectionInterface;
 use Sonata\Component\Product\DeliveryInterface;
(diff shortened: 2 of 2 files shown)


greg0ire:
Some tests fail because the Pool is now `final`. It should not be mocked, the real object should be used instead.

composer.json
```diff
@@ -27,7 +27,7 @@
         "sonata-project/core-bundle": "^3.12.0",
         "sonata-project/datagrid-bundle": "^2.3",
         "sonata-project/doctrine-orm-admin-bundle": "^3.2",
-        "sonata-project/formatter-bundle": "^3.4",
+        "sonata-project/formatter-bundle": "^3.4 || 4.0",
```
phansys: Trust sem ver:
```suggestion
        "sonata-project/formatter-bundle": "^3.4 || ^4.0",
```

tests/ProductBundle/Controller/Api/ProductControllerTest.php
```diff
@@ -302,9 +303,17 @@ public function createProductController($product = null, $productManager = null,
...
 
         return new ProductController($productManager, $productPool, $formFactory, $formatterPool);
     }
+    
+    
+    private function getFormatterPool()
+    {
+        $pool = new FormatterPool('whatever');
+        $pool->setLogger($this->createMock(LoggerInterface::class));
+        return $pool;
```
phansys:
```suggestion

        return $pool;
```

tests/ProductBundle/Controller/Api/ProductControllerTest.php
```diff
@@ -302,9 +303,17 @@ public function createProductController($product = null, $productManager = null,
...
         if (null === $formatterPool) {
-            $formatterPool = $this->createMock(FormatterPool::class);
+            $formatterPool = $this->getFormatterPool();
         }
 
         return new ProductController($productManager, $productPool, $formFactory, $formatterPool);
     }
+    
+    
+    private function getFormatterPool()
```
phansys:
```suggestion
    private function getFormatterPool(): FormatterPool
```

tests/ProductBundle/Controller/Api/ProductControllerTest.php
```diff
@@ -302,9 +303,17 @@ public function createProductController($product = null, $productManager = null,
...
         }
         if (null === $formatterPool) {
-            $formatterPool = $this->createMock(FormatterPool::class);
+            $formatterPool = $this->getFormatterPool();
         }
 
         return new ProductController($productManager, $productPool, $formFactory, $formatterPool);
     }
+    
+
```
phansys:
```suggestion
```

tests/ProductBundle/Controller/Api/ProductControllerTest.php
```diff
@@ -143,7 +144,7 @@ public function testPostProductAction(): void
         $productPool = $this->createMock(ProductPool::class);
         $productPool->expects($this->once())->method('getManager')->will($this->returnValue($productManager));
 
-        $formatterPool = $this->createMock(FormatterPool::class);
+        $formatterPool = $this->getFormatterPool();
         $formatterPool->expects($this->exactly(2))->method('transform');
```
phansys: `$formatterPool` isn't a mock instance now.

quisse: I'm not quite familiar with unit tests. AFAIK this mock is used to determine if the transform method is called twice. But since this shouldn't be a mock anymore, is this check still necessary/possible?

phansys: The test checking if the method is called is not required anymore since you're using the real object now, so the call to `transform()` (or any other method) will be really performed and the logic from that class will be responsible to produce a valid object.
The goal of the count on the mock was to ensure the involved methods of that class were called, because all the class methods were mocked too.

quisse: Ok, thanks!

core23:
Please approve @phansys

PR #575, opened 2019-03-10
Use new datepicker

quisse:
## Subject

I am targeting this branch, because not backwards compatible.

## Changelog

```markdown
### Changed
- use newer datepicker
```

### DIFF
--- a/src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php
+++ b/src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php
@@ -17,14 +17,26 @@ use Sonata\EasyExtendsBundle\Mapper\DoctrineCollector;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 
 /**
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-class SonataCustomerExtension extends Extension
+class SonataCustomerExtension extends Extension implements PrependExtensionInterface
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function prepend(ContainerBuilder $container): void
+    {
+        if ($container->hasExtension('twig')) {
+            // add custom form widgets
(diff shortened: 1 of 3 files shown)


src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php
```diff
@@ -18,13 +18,24 @@
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
```
pierredup: This needs to be loaded in alphabetical order

```diff
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
-use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
```

src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php
```diff
@@ -17,14 +17,26 @@
...
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function prepend(ContainerBuilder $container): void
+    {
+        if ($container->hasExtension('twig')) {
+            // add custom form widgets
+            $container->prependExtensionConfig('twig', ['form_themes' => ['@SonataCore/Form/datepicker.html.twig']]);
+        }
```
greg0ire: Why is the PR not backwards compatible and why is this necessary now, and was not before?

quisse: Will look into this

quisse: I guess it ís backwards compatible (I've rebased). 
Without this, the input used is just a normal text input. Am I missing something or shall i add this fragment to `Sonata\OrderBundle\DependencyInjection\SonataOrderExtension` so the bundles can be used separately?

greg0ire: You can do that yeah. Since it is BC, maybe contribute it to the stable branch then?

quisse: Was a bit confused but had to change the base branch here. (instead of rebasing in git 🤦‍♂)

greg0ire: I think you would have to do both things :P

greg0ire:
Thanks @quisse !

PR #581, opened 2019-03-15
fixed some relations

quisse:
## Subject

I am targeting this branch, because backwards compatible.

## Changelog

```markdown
### Fixed
Missing relations
```

### DIFF
--- a/src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php
+++ b/src/CustomerBundle/DependencyInjection/SonataCustomerExtension.php
@@ -126,6 +126,16 @@ class SonataCustomerExtension extends Extension
             'orphanRemoval' => false,
         ]);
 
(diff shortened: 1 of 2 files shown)


greg0ire:
What is broken and how was this relationship missing? All relationships need not be bidirectional. In fact, bidirectional relationship are not precisely recommended.

quisse:
@greg0ire based on suggestions from the Symfony profiler... The profiler shows following for example: 
Untitled
So to me it seems like that there's something wrong. Maybe I get this wrong?

greg0ire:
Oh indeed, you can't really have unidirectional 1-to-n relationships: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/association-mapping.html#one-to-many-unidirectional-with-join-table, I think your fix is easiest way to get us closer to something correct.

OskarStark:
Thank you

quisse:
Great! ☺️

PR #584, opened 2019-03-24
Be more assertive

greg0ire:
## Subject

Saying a method can be called any time is saying it could be in
particular becalled zero times. PHPunit seems not to take calls to any()
into account, and rightfully so. That means the test is marked as risky.
I replaced these calls with calls leading to more precise expectations,
eliminating the only risky test in the test suite at the moment.

I am targeting this branch, because this is BC (I touched only tests).


OskarStark:
Thank you @greg0ire

PR #600, opened 2019-10-04
Master - ProductBundle - formatCurrency() expects parameter 2 to be string, 

salamichel:
Subject : 
ProductBundle issue on front page:
- NumberFormatter::formatCurrency() expects parameter 2 to be string, object given
- Error in VariationsFormBlockService  - optionResolver get method not available

## Changelog
- number_format_currency accept string not currencyInterface 
- correction VariationsFormBlockService use method offsetGet instead Get

PR #604, opened 2019-10-09
master - twitter -use getLabel from currency

salamichel:
related to #600


greg0ire:
Why are there 3 commits to change one line?

PR #653, opened 2020-02-13
Remove mapping from product collection to generic Collection Entity

### LINKED ISSUE #644: Missing property error ($productCollection) caused by the ProductExtension
#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
# Put the result here.
```
sonata-project/classification-bundle     3.9.2  3.10.0 Symfony SonataClassificationBundle
sonata-project/ecommerce                 3.2.2  3.2.2  E-Commerce solution provided by Sonata

## Subject
In this pull request: https://github.com/sonata-project/ecommerce/pull/585
@quisse added a fix for a missing relation. However, if I understand it correctly this requires you to add the property "productCollection" to your Collection entity.
```
$collector->addAssociation($config['class']['collection'], 'mapOneToMany', [
      'fieldName' => 'productCollection',
      'targetEntity' => $config['class']['product_collection'],
      'cascade' => [
          'persist',
      ],
      'mappedBy' => 'collection',
      'orphanRemoval' => false,
]);
``` 
In our project we use the Collection entity of the Classification bundle for multiple purposes. So we would rather not add a productCollection property to the generic Collection Entity. 

Did we configure it incorrectly? Or should this association be optional?

jeroendk:
### Remove mapping to Classification\Collection to stop forcing people to add an extra property to their generic Collection class. 

I am targeting this branch, because (I think) this is a bugfix. not sure though... Some people might actually rely on the association being there? For our use case, it is a bug. 

Another option would be to check if the productCollection property exists in the Collection class and if so add the association? We could remove the association altogether in a new major? I would like one's input on this. 

Closes #644

## Changelog

```markdown

### Fixed
- Only add association mapping from ProductCollection to Collection when property exists
- Only add inversed side definition in ProductionCollection mapping when property exists. 
```

### DIFF
--- a/src/ProductBundle/DependencyInjection/SonataProductExtension.php
+++ b/src/ProductBundle/DependencyInjection/SonataProductExtension.php
@@ -220,7 +220,7 @@ class SonataProductExtension extends Extension
              'orphanRemoval' => false,
         ]);
 
-        if ($addCollectionAssociation){
+        if ($addCollectionAssociation) {
             $collector->addAssociation($config['class']['collection'], 'mapOneToMany', [
                 'fieldName' => 'productCollection',
                 'targetEntity' => $config['class']['product_collection'],
--- a/src/ProductBundle/DependencyInjection/SonataProductExtension.php
+++ b/src/ProductBundle/DependencyInjection/SonataProductExtension.php
@@ -192,6 +192,11 @@ class SonataProductExtension extends Extension
         ]);
 
         $addCollectionAssociation = false;
+
+        /**
+         * NEXT_MAJOR: remove this check, the inverse declaration below
+         * and association added to the Collection entity
+         */
         if(\property_exists($config['class']['collection'], 'productCollection')){
             $addCollectionAssociation = true;
         }
--- a/src/ProductBundle/DependencyInjection/SonataProductExtension.php
+++ b/src/ProductBundle/DependencyInjection/SonataProductExtension.php
@@ -193,11 +193,11 @@ class SonataProductExtension extends Extension
 
         /**
          * NEXT_MAJOR: remove this check, the inverse declaration below
-         * and association added to the Collection entity
+         * and association added to the Collection entity.
          */
         $addCollectionAssociation = false;
-        
-        if(\property_exists($config['class']['collection'], 'productCollection')){
+
+        if (property_exists($config['class']['collection'], 'productCollection')) {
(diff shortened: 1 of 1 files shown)


src/ProductBundle/DependencyInjection/SonataProductExtension.php
```diff
@@ -210,16 +209,6 @@ public function registerDoctrineMapping(array $config): void
...
 
-        $collector->addAssociation($config['class']['collection'], 'mapOneToMany', [
-            'fieldName' => 'productCollection',
-            'targetEntity' => $config['class']['product_collection'],
-            'cascade' => [
-                'persist',
-            ],
-            'mappedBy' => 'collection',
-            'orphanRemoval' => false,
-        ]);
```
greg0ire: Would it make sense to do this conditionally, if `property_exists($config['class']['collection'], 'productCollection')`? I did not follow the whole issue, but I since you say this fix might actually break people's stuff…

jeroendk: Well, that is what I proposed, to add it conditionally. In theory, it could break if someone relies on the association being there right? 

I will update this PR to add it conditionally but propose to remove it altogether in the master branch.

greg0ire: > I will update this PR to add it conditionally but propose to remove it altogether in the master branch.

`[this is the way](https://media1.tenor.com/images/614de537ea516c274f9c95e722fbf9d5/tenor.gif?itemid=15630699)

greg0ire:
Thanks @jeroendk !

jeroendk:
@greg0ire What is your release cycle for the Ecommerce bundle / when can I expect a release with this fix?

wbloszyk:
@jeroendk 
Can u look at this  [sandbox#620](https://github.com/sonata-project/sandbox/pull/620#discussion_r383779659) ? 
Also can you check the same thing like here with User-Customers relation?

jeroendk:
@wbloszyk sure, where can I find the User-Customers relation (discussion)?

greg0ire:
> @greg0ire What is your release cycle for the Ecommerce bundle / when can I expect a release with this fix?

People like you ask me and I release. Expect one for tonight.

jeroendk:
> People like you ask me and I release. Expect one for tonight.

Awesome, thanks!

PR #671, opened 2020-05-20
Fix mssql limit issue

wbloszyk:
## Subject

This PR replace limit from database to php. This will resolve problem with diffrence usages of limit function in some sql. 
#### SQL Server / MS Access Syntax:
```sql
SELECT TOP number|percent column_name(s)
FROM table_name
WHERE condition;
```
#### MySQL Syntax:
```sql
SELECT column_name(s)
FROM table_name
WHERE condition
LIMIT number;
```

#### Oracle Syntax:
```sql
SELECT column_name(s)
FROM table_name
WHERE ROWNUM <= number;
```

I am targeting this branch, because it is PATCH.

Part of https://github.com/sonata-project/dev-kit/issues/689

### DIFF
--- a/src/ProductBundle/Entity/ProductCategoryManager.php
+++ b/src/ProductBundle/Entity/ProductCategoryManager.php
@@ -101,10 +101,9 @@ class ProductCategoryManager extends BaseEntityManager implements ProductCategor
                 AND (c.enabled = :categoryEnabled OR c.enabled IS NULL)
                 AND p.parent_id IS NULL
                 AND pc.category_id = :categoryId
-                LIMIT %d
                 ) AS cnt';
 
-        $sql = sprintf($sql, $metadata->table['name'], $productMetadata->table['name'], $categoryMetadata->table['name'], $productMetadata->table['name'], $limit);
+        $sql = sprintf($sql, $metadata->table['name'], $productMetadata->table['name'], $categoryMetadata->table['name'], $productMetadata->table['name']);
 
         $statement = $this->getConnection()->prepare($sql);
         $statement->bindValue('productEnabled', 1);
@@ -113,9 +112,10 @@ class ProductCategoryManager extends BaseEntityManager implements ProductCategor
         $statement->bindValue('categoryId', $category->getId());
 
         $statement->execute();
-        $res = $statement->fetchAll();
+        $result = $statement->fetchAll();
+        $count = $result[0]['cntId'];
 
-        return $res[0]['cntId'];
+        return $limit > $count ? $count : $limit;
     }
 
     /**


greg0ire:
What is the issue? How does your patch fix it?

src/ProductBundle/Entity/ProductCategoryManager.php
```diff
@@ -101,10 +101,9 @@ public function getProductCount(CategoryInterface $category, $limit = 1000)
                 AND (c.enabled = :categoryEnabled OR c.enabled IS NULL)
                 AND p.parent_id IS NULL
                 AND pc.category_id = :categoryId
-                LIMIT %d
```
greg0ire, line 104: Have you seen the comment on line 87? Also, what are the consequences of removing this `LIMIT` for other RDBMSs?

wbloszyk: I update RP description.

phansys: IMO, we should try to leverage the DBAL as much as possible in order to avoid dealing with specific RDBMSs by our own responsibility. This way, we could build the subquery using DQL or a query builder, and then if it is really required, get the raw SQL query to be used with the main query.
In the worst case, if that is not possible, we should build the different syntaxes by hand.
Even if this query is just intended to return the counted results, I guess leaving the limit open in the persistence layer is not good.

greg0ire: @phansys have you seen the comment on line 87? @wbloszyk what about you?

phansys: > @phansys have you seen the comment on line 87? @wbloszyk what about you?

Yes. What I mean is something like this:
```
$dql = 'SELECT DISTINCT(pc.product_id) FROM %s pc ... LIMIT %d';
$sql = 'SELECT count(cnt.product_id) AS "cntId" FROM('.$dql->getSQL().');';
```

greg0ire: Oh right, I read to fast it seems… sorry!

wbloszyk: > > @phansys have you seen the comment on line 87? @wbloszyk what about you?
> 
> Yes. What I mean is something like this:
> 
> ```
> $dql = 'SELECT DISTINCT(pc.product_id) FROM %s pc ... LIMIT %d';
> $sql = 'SELECT count(cnt.product_id) AS "cntId" FROM('.$dql->getSQL().');';
> ```

- Subqueris is not allowed.
- Generating SQL crash (DQL look the same)
- Modify SQL for each platform is to mest

`ProductCategoryManager::getProductCount()` is only use in `ProductMenuBuilder::getCategoryTitle()`  to display counts in menu.

We can inject ProductManager and use:
https://github.com/sonata-project/ecommerce/blob/9c2a75ce27e2070eeb75349f2d542e49a0368756/src/ProductBundle/Entity/ProductManager.php#L65

This change will also limit products count when filters will be set.

phansys: AFAIK, the syntax `SELECT COUNT(DISTINCT pc.product_id)` is [supported by DQL](https://www.doctrine-project.org/projects/doctrine-orm/en/2.5/changelog/migration_2_5.html#dql-improvements), even by query builder with `Expr::countDistinct()`. Why we need a subquery then?

wbloszyk: I don't know doctrine to much. Anyway we should get the same result from `ProductCategoryManager::getProductCount()` and `ProductManager::getProductCountInCategory` 
Names are example. We should reuse query as much as possible. It is also yours words. I will add some core example in Monday.

phansys: Great! :+1:
Just let me know if you need help. I'm not using this bundle, but I can provide support with the ORM if required.

PR #674, opened 2020-06-02
Drop support for Symfony < 4.3

wbloszyk:
## Subject

Drop support for Symfony < 4.3.


src/BasketBundle/Controller/BasketController.php
```diff
@@ -34,7 +35,7 @@
 /**
  * This controller manages the Basket operation and most of the order process.
  */
-class BasketController extends Controller
+class BasketController extends AbstractController
```
jordisala1991, line 38: This look like a BC break

jordisala1991:
Can you fill the PR template?

src/BasketBundle/Controller/BasketController.php
wbloszyk: I think i should do it in master

wbloszyk:
Should i drop symfony 4.3 too?

jordisala1991:
Is it worth it?

wbloszyk:
Symfony 4.3 is not maintained. I think we can keep support for 4.3 and drop in in master.

jordisala1991:
That would be good I think. Make all BC changes on 3.x. Upgrade to 4.4 and bc break on master

jordisala1991:
Thank you @wbloszyk

wbloszyk:
@core23 I know but diffrence is not do much. Also all sonata 3 bundles support it. It will be good idea to keep minimum require symfony 4.3 in whole sonata 3.

core23:
What I wanted to say is that our release cycle is very slow for this bundle. IMHO you can already create a new PR to drop 4.3.

wbloszyk:
I will focus on sandbox and PageBundle now. I needed in to my 3 commerce project. I have many improve in my roadmap for this bundle too. I will update some issue on next week like sonata 4, blocks upgrade, sandbox 3 and 4 roadmap, ecommerce.

greg0ire:
Should this be patch or minor?

wbloszyk:
Next release should be minor and should include https://github.com/sonata-project/ecommerce/pull/678. This PR is RTM. Will be awesome release it with drop CoreBundle (both drop `twig/extensions`).

PR #672, opened 2020-05-20
Make CoreBundle optional

wbloszyk:
## Subject

I am targeting this branch, because this changes are BC.

Part of https://github.com/sonata-project/dev-kit/issues/697

## Changelog

```markdown
### Removed
- remove SonataCoreBundle dependencies
```

PR #675, opened 2020-06-11
Update composer.json

wbloszyk:
## Subject

This change will allow install optionally only CoreBundle v3.20 (which constains all deprecations notes).

#### Remove process will be look like:
- update CoreBundle to 3.20
- remove deprecations
- remove CoreBundle
- upgrade extensions to 1.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,8 @@
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.2 || >=3.0",
-        "jms/serializer": "<0.13"
+        "jms/serializer": "<0.13",
+        "sonata-project/core-bundle": "<3.20"
     },
     "require-dev": {
         "friendsofsymfony/rest-bundle": "^2.2",


jordisala1991:
Thank you @wbloszyk

greg0ire:
Please add a changelog

wbloszyk:
What should I add?

greg0ire:
Something directed at end users. Keep in mind that users probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
Please make it compliant with http://keepachangelog.com

wbloszyk:
@greg0ire It is part of https://github.com/sonata-project/ecommerce/pull/672. I Should add it there. I don't think changelog here is require becouse previus PR is not release. WDYT? I also don't know what should i write on this changelog too.

greg0ire:
Oh right I just had a look at the changes, not worthy of a changelog note indeed, let's forget it.

PR #676, opened 2020-06-16
Fix dispatch event to work with contracts

wbloszyk:
## Subject

This PR replace event dispach parameter to work correctly with constracts.

## Changelog

```markdown
### Fixed
- Deprecations for event dispatching
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
         "symfony/dependency-injection": "^4.3",
         "symfony/doctrine-bridge": "^4.3",
         "symfony/event-dispatcher": "^4.3",
+        "symfony/event-dispatcher-contracts": "^1.1 || ^2.0",
         "symfony/filesystem": "^4.3",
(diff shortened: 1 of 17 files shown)


jordisala1991:
You should change: `use Symfony\Component\EventDispatcher\EventDispatcherInterface;` to use the contract.

wbloszyk:
@jordisala1991 
Done. I copy it from AdminBundle. For now version ^2.0 will not be use becouse symfony/event-dispatcher require ^1.1. Anyway support for Symfony5 is close so i keep it here too.

wbloszyk:
@core23 I update test too. Should I do something more?

core23:
Please add a changelog entry, after that it's RTM

wbloszyk:
@core23 Changelog can be like this?

core23:
Thanks @wbloszyk

PR #678, opened 2020-06-22
Replace truncate filter with u filter

wbloszyk:
## Subject

This bundle was using `truncate` filter without explicitly require `twig/extensions`, with this PR it uses `u` filter.

I added `twig/extra-bundle` to autoload the Twig extension and a fallback in case Flex is not used.

Ref: sonata-project/SonataAdminBundle#6132

I am targeting this branch, because these changes are BC.

## Changelog

```markdown
### Added
- Added `twig/string-extra` dependency.
### Changed
- Changed use of `truncate` filter with `u` filter.
```

### DIFF
--- a/src/ProductBundle/DependencyInjection/SonataProductExtension.php
+++ b/src/ProductBundle/DependencyInjection/SonataProductExtension.php
@@ -17,8 +17,10 @@ use Sonata\EasyExtendsBundle\Mapper\DoctrineCollector;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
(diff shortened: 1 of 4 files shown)


wbloszyk:
RTM

src/ProductBundle/DependencyInjection/SonataProductExtension.php
```diff
@@ -348,4 +352,14 @@ protected function registerSeoParameters(ContainerBuilder $container, array $con
             $container->setParameter(sprintf('sonata.product.seo.product.%s', $key), $value);
         }
     }
+
+    private function configureStringExtension(ContainerBuilder $container): void
+    {
+        if (!$container->hasDefinition('twig.extension.string') || !is_a($container->getDefinition('twig.extension.string')->getClass(), StringExtension::class)) {
+            $definition = new Definition(StringExtension::class);
```
core23, line 359: Will this work if our other bundles will register the same `StringExtension::class` service?

wbloszyk: It should work, becouse result will be the same.

phansys:
Thank you @wbloszyk!

wbloszyk:
Thank you @franmomu

PR #687, opened 2020-07-21
Translate flash messages

wbloszyk:
## Subject

Translator dependencies was remove in `twig-extensions 1.x`. Messages should be translate before add it to `flash bag`.

## Changelog

```markdown
### Changed
- Translate flash message before add it to `flash bag`
```


wbloszyk:
It is RTM

PR #688, opened 2020-07-24
Allow "friendsofsymfony/rest-bundle:^3.0"

wbloszyk:
## Subject

Allow "friendsofsymfony/rest-bundle:^3.0".

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Added
- support for "friendsofsymfony/rest-bundle:^3.0"
### Change
- support for deprecated "rest" routing type in favor for xml
```

### DIFF
--- a/src/BasketBundle/Controller/Api/BasketController.php
+++ b/src/BasketBundle/Controller/Api/BasketController.php
@@ -14,10 +14,9 @@ declare(strict_types=1);
 namespace Sonata\BasketBundle\Controller\Api;
 
 use FOS\RestBundle\Context\Context;
-use FOS\RestBundle\Controller\Annotations\QueryParam;
-use FOS\RestBundle\Controller\Annotations\View;
+use FOS\RestBundle\Controller\Annotations as Rest;
 use FOS\RestBundle\Request\ParamFetcherInterface;
-use FOS\RestBundle\View\View as FOSRestView;
+use FOS\RestBundle\View\View;
 use Nelmio\ApiDocBundle\Annotation\ApiDoc;
 use Sonata\Component\Basket\BasketBuilderInterface;
 use Sonata\Component\Basket\BasketElementInterface;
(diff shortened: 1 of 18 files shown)


src/BasketBundle/Controller/Api/BasketController.php
```diff
@@ -14,8 +14,7 @@
 namespace Sonata\BasketBundle\Controller\Api;
 
 use FOS\RestBundle\Context\Context;
-use FOS\RestBundle\Controller\Annotations\QueryParam;
-use FOS\RestBundle\Controller\Annotations\View;
+use FOS\RestBundle\Controller\Annotations as Rest;
 use FOS\RestBundle\Request\ParamFetcherInterface;
 use FOS\RestBundle\View\View as FOSRestView;
```
phansys, line 19: The alias is not required anymore.

src/BasketBundle/Controller/Api/BasketController.php
```diff
@@ -134,9 +134,9 @@ public function getBasketsAction(ParamFetcherInterface $paramFetcher)
      *  }
      * )
      *
-     * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
+     * @Rest\View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
-     * @param $id
+     * @param int $id Basket identifier
```
phansys, line 139: Is the model forcing the basket to allow only integer ids?

wbloszyk: Model allow for `mixed`. I use `int` to be unequivocal with `@ApiDoc`

phansys: There is an issue with the `@ApiDoc` annotation then. I think we should not propagate the same problem here.
I understand you're trying to be consistent, but in this case the reference type does not seem to be the right choice.

wbloszyk: I reverted it (docblock type hints). Lets do it in seperate PR.

wbloszyk:
@sonata-project/contributors Can you check it?

wbloszyk:
I will add it at the end, after add support for nelmio/api-doc-bundle v3

wbloszyk:
@OskarStark Can you look at DOCtor-RTS? You make somethig simmilar ([to this](https://github.com/sonata-project/ecommerce/pull/701)), where nelmio/api-doc work with `.yml`.

PR #771, opened 2021-02-07
Fix support for doctrine/doctrine-bundle v2

wbloszyk:
## Subject

## Changelog

```markdown
### Changed
- Change usage of `Symfony\Bridge\Doctrine\RegistryInterface` into `Doctrine\Persistence\ManagerRegistry` (BC)
```

### DIFF
--- a/src/Component/Product/SeoProductIterator.php
+++ b/src/Component/Product/SeoProductIterator.php
@@ -13,10 +13,10 @@ declare(strict_types=1);
 
 namespace Sonata\Component\Product;
 
+use Doctrine\Persistence\ManagerRegistry;
 use Sonata\Exporter\Source\DoctrineDBALConnectionSourceIterator;
 use Sonata\Exporter\Source\SourceIteratorInterface;
 use Sonata\Exporter\Source\SymfonySitemapSourceIterator;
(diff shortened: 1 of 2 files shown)


dmaicher (approved):

VincentLanglet (commented):
Can't we make this work for both ^1 and ^2 ?

VincentLanglet (approved):

wbloszyk:
@VincentLanglet We can keep ^1 and it will be work for both. Anyway in sandbox (so all ower bundles) we can use ^2. 

We should keep support for 1.x or we can drop support for it becouse we fully support 2.x? WDYT @sonata-project/contributors @greg0ire

greg0ire:
I'm torn: if it means extra work to keep ^1 I wouldn't keep it, but you seem to say it wouldn't. If there are code cleanups we could do by dropping it, then I would drop it.

wbloszyk:
It is RTM.

VincentLanglet (approved):

dmaicher (approved):

OskarStark (approved):

OskarStark:
Thanks

